PR #433, opened 2026-03-04
Tests and implementation for `preprocess_implicit_and`

jbylund:
Adds a dedicated, parameterized test suite for `preprocess_implicit_and` and reimplements the function so that implicit AND is converted to explicit AND in search queries using a tokenizer aligned with the main parser grammar.

## Motivation

- **Test coverage**: `preprocess_implicit_and` was only exercised indirectly via `test_parse_implicit_and` (full parse of `"a b"`). There were no direct tests for the preprocessor’s input → output behavior.
- **Spec as tests**: The new tests act as the specification (before/after query pairs), so the implementation can be validated and changed safely.
- **Implementation**: The previous implementation was removed and replaced with a tokenizer-based approach that reuses the same conceptual primitives as the main grammar (quoted strings, regex, parens, keywords, comparison/arithmetic ops, words, mana, etc.), then inserts AND and serializes without extra spaces.

### New file: `api/parsing/tests/test_preprocess_implicit_and.py`

- **Parameterized tests** over a single `TESTCASES` list of `{query, expected, id}` dicts (~68 cases).
- **Happy path**: adjacent words, single tokens, explicit AND/OR, attribute pairs, parentheses, quoted strings (single/double), regex, comparison operators, arithmetic (e.g. `power+toughness>cmc+cmc`), negation (`a -b`, `-t:creature`, `id=r -o:enchantment`), hyphenated words and values (`some-word`, `otag:40k-model`, `old-growth-troll`, `otag:dual-land`, `is:modal-dfc`, etc.), mana (`c:{w}{u}`, `c:{W/U}`), date/year (`date:2025`, `year:2024`), whitespace normalization, empty input.
- **Error cases**: unclosed double quote, unclosed single quote, unclosed regex → `ValueError` with message containing `"Unmatched"`.
- **Output format**: No extra whitespace (e.g. `cmc=3` stays `cmc=3`, `(a b)` → `(a AND b)` not `( a AND b )`).

### Modified: `api/parsing/parsing_f.py`

- **`_get_implicit_and_tokenizer()`**: Builds a pyparsing tokenizer that returns a list of raw token strings. Uses the same token shapes as the main grammar (quoted strings, `/.../` regex, `( )`, AND/OR, comparison/arithmetic ops, float, string/value word, mana pattern). Token order is chosen so that e.g. `"bar"` and `"bolt"` are one token (string_value before mana).
- **`_tokenize_for_implicit_and(query)`**: Runs the tokenizer, returns token list; on `ParseException` raises `ValueError("Unmatched quote or regex in query")`. Detects unclosed regex (leading `"/"` with no second `"/"`) and raises `ValueError("Unmatched / in regex pattern in query")`.
- **`preprocess_implicit_and(query)`**: Tokenizes, inserts `AND` between consecutive operands (and between operand and `-`, and after `)` / before `(` where appropriate), then serializes by concatenating tokens with no space except `" AND "` and `" OR "` for those keywords. Empty or whitespace-only input returns `""`.
- **`is_operator()`**: Unchanged; still used by the AND-insertion logic.

## Testing

- All 68 tests in `test_preprocess_implicit_and.py` pass.
- `test_parse_implicit_and` in `test_parsing.py` still passes (full pipeline with the new preprocess).

## Notes

- The tokenizer is built from standalone parser elements (not imported from `create_basic_parsers()` etc.) but mirrors the same token boundaries and ordering so preprocess output is consistent with what the main parser expects.
- Hyphenated and “otag”-style cases (e.g. `otag:40k-model`, `otag:dual-land`, `old-growth-troll`) are included so that real query patterns used elsewhere in the codebase are covered.

### DIFF
--- a/api/parsing/parsing_f.py
+++ b/api/parsing/parsing_f.py
@@ -807,7 +807,12 @@ def _get_implicit_and_tokenizer() -> ParserElement:
     # the regex requires the last character to be alphanumeric/underscore.
     string_value_tok = Regex(r"[a-zA-Z0-9_]([a-zA-Z0-9_-]*[a-zA-Z0-9_])?").setParseAction(lambda t: t[0])
 
-    # Mana pattern (e.g. {1}{G}, {w}{u}) as one token; only after word so "bar" isn't "b" + "ar"
+    # Mana pattern (e.g. {1}{G}, {w}{u}, {2/W}G) as one token.
+    # mana_tok only fires when the sequence STARTS with "{" (string_value_tok above would
+    # have matched otherwise).  Once inside OneOrMore, simple_mana_symbol can absorb bare
+    # mana chars (G, 1, …) that follow a curly symbol, e.g. {2/W}G → one token.
+    # string_value_tok must precede mana_tok in one_token so "bar" stays one word, not
+    # "b" (simple mana) + "ar".
     curly_mana_symbol = Regex(r"\{[^}]+\}")
     simple_mana_symbol = Regex(r"[0-9WUBRGCXYZwubrgcxyz]")
     mana_tok = Combine(OneOrMore(curly_mana_symbol | simple_mana_symbol)).setParseAction(lambda t: t[0])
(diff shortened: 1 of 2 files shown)


jbylund:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/jbylund/arcane_tutor/pull/433#pullrequestreview-3893170825)

But for the one just do `t.lower() in ...`

PR #548, opened 2026-06-25
Shared Memory Cache

jbylund:
## Problem

Each worker process maintains its own independent `LRUCache`. A response cached by worker A is a miss for worker B, so under round-robin load balancing you see hits and misses alternating indefinitely — the effective hit rate is `1 - (1/N)^k` where N is worker count and k is request count per worker, which is terrible at startup.

## Solution

A new `shared_cache` Rust extension backs `CachingMiddleware` with a memory-mapped file shared across all worker processes. The first worker to cache a response makes it a hit for every other worker immediately.

### shared_cache design

- **Generational ring buffer**: N equal-sized pages. One active page accepts writes; the remaining N-1 sealed pages serve lock-free reads. When the active page fills, a two-phase rotation runs: a lock-free survivor scan (SIEVE-style `visited` bit), then an O(1) locked commit.
- **Locking model**: spinlock only for the active page and filter inserts. Sealed page reads are lock-free — a per-page generation counter detects concurrent rotation.
- **Cuckoo filter**: 16-bit fingerprints, 4 slots/bucket. True misses skip the spinlock entirely (~31 ns).
- **set() fast path**: before rkyv serialization, checks filter → body length → sampled content hash (xxh3 of first+last 64 bytes). If all match, the write is skipped (~536 ns vs ~1155 ns for a full insert).
- **`pop(key)`**: runs entirely under the spinlock. Active page: decrements entry_count, deletes from filter, writes TOMBSTONE. Sealed pages: brackets the TOMBSTONE write with generation bumps (even→odd→even) so the lock-free `scan_survivors()` detects concurrent tombstones via its per-chunk seqlock and retries only the affected chunk rather than discarding the whole page.
- **`invalidate()`**: implemented on `CachingMiddleware` and delegates to `SharedCache.invalidate()`. Wiring it into `APIResource._clear_caches()` so bulk imports flush the shared cache is deferred to a follow-up PR.

### CachingMiddleware changes

- Cache key type changed from `tuple` to `bytes` (`orjson.dumps`) — needed for the Rust layer; also faster than Python tuple hashing.
- `CachedResponse.headers` changed from `dict[str, str]` to `list[tuple[str, str]]` — preserves insertion order and is the format SharedCache deserializes to.
- `cacheable_headers()` updated to match.
- Added `_CacheProtocol` so any duck-typed cache (LRUCache or SharedCache) is accepted without `Any`.
- Added `invalidate()` method that delegates to the inner cache.
- Added PID to log messages so hit/miss patterns across workers are visible.
- Cache key serialized once in `process_request` and reused in `process_response`.

### Wiring

`api_worker.py` opens `SharedCache` post-fork (so each process gets its own file descriptor to the shared mmap). Falls back to per-process LRUCache on `ImportError` or `OSError`. Path is configurable via `SHARED_CACHE_PATH` env var; blue and green use distinct paths.

### Docker

Added `shared_cache` to the rust-builder stage in `api/Dockerfile` — both `card_engine` and `shared_cache` wheels are built there and installed in the app stage.

### CI

`unit-tests.yml` now builds and caches a `shared_cache` wheel alongside `card_engine`; the Rust toolchain and Cargo cache steps gate on either wheel being stale. `labeler.yml` adds `rust` and `shared_cache` labels (and renames the old `engine` label to `card_engine`).

## Performance

Measured on Apple M-series, 5,025-byte gzip response body, 500 warm keys:

| Operation | Latency |
|---|---|
| get hit (full middleware path) | ~364 ns |
| get miss (cuckoo filter short-circuit) | ~31 ns |
| set (same content, fast path) | ~536 ns |
| set (new/changed content) | ~1,155 ns |

The miss path is on par with an in-process LRUCache miss (because it short circuits after consulting the cuckoo filter). The hit path is slower than a local LRU hit (at ~51 ns) due to rkyv deserialization and mmap→PyBytes copy, but the cross-process sharing means the total cache pool is 1× regardless of worker count instead of N×.

**Redis baseline:** a cross-container `GET` of the same 5,025-byte value benchmarks at p50 ~62 µs, p95 ~79 µs (10,000 iterations, Python `redis-py` client, both containers on the same Docker bridge network). That is ~170× slower than the shared-memory hit path. Redis is the obvious alternative for cross-process sharing, but the loopback round-trip cost alone makes it the wrong tool when the data fits in memory on the same host.


shared_cache/src/region.rs
```diff
@@ -0,0 +1,159 @@
...
+    ((maxsize * 4).next_power_of_two() / 4).max(16)
+}
+
+// ── Atomic helpers ─────────────────────────────────────────────────────────────
+
+/// Atomic read of the page's generation counter (Acquire).
+pub fn read_page_generation(base: *const u8) -> u32 {
+    // generation is at offset 12 within PageHeader
+    let ptr = unsafe { base.add(12) as *const AtomicU32 };
+    fence(Ordering::Acquire);
```
jbylund: **Bug (UB on ARM): `fence(Acquire)` is before the load, not after.**

The C++/Rust memory model requires the Relaxed load to be *sequenced before* the acquire fence for it to synchronize-with a Release store. The code has the order reversed:

```rust
fence(Ordering::Acquire);            // ← before
(*ptr).load(Ordering::Relaxed)       // ← load comes after
```

On x86/TSO this is accidentally correct (loads carry implicit acquire semantics in hardware). On aarch64 (`dmb ishld` then `ldr`) the CPU is free to hoist the load above the barrier. Result: `gen_before` and `gen_after` in `get_with()` and `pop()` can both see the same stale pre-rotation value → generation check passes → `rkyv::access_unchecked` runs on zeroed/partially-overwritten mmap bytes → UB. The benchmarks were measured on Apple M-series, so this is live.

Fix: swap the two lines, or use `load(Ordering::Acquire)` and drop the separate fence:
```rust
pub fn read_page_generation(base: *const u8) -> u32 {
    let ptr = unsafe { base.add(12) as *const AtomicU32 };
    unsafe { (*ptr).load(Ordering::Acquire) }
}
```

shared_cache/src/gen_cache.rs
```diff
@@ -0,0 +1,695 @@
...
+            return None;
+        }
+        let active_idx = self.active_idx();
+        let active_abs_len = self.do_probe(active_idx, hash, key).map(|(off, len, _)| {
+            (self.page_offset(active_idx) + self.arena_start_in_page + off as usize, len as usize)
+        });
+        unlock(&self.mmap);
+
+        if let Some((abs, len)) = active_abs_len {
+            return Some(f(&self.mmap[abs..abs + len]));
```
jbylund: **Bug (UB): active-page arena is read after the lock is released with no generation guard.**

After snapshotting `(abs, len)` and calling `unlock()`, another worker can run two full rotations (seal page A, zero page B; seal page B, zero page A). The bytes at `abs..abs+len` are now zeroed, and the call to `f` (which feeds into `rkyv::access_unchecked`) is UB.

With `n_pages=2` — the **default** in the PyO3 signature (`#[pyo3(signature = (..., n_pages=2))]`) — only two rotations are needed. The sealed-page loop already has the correct gen-before/gen-after guard; the active-page branch needs the same:

```rust
let gen_before = self.page_generation(active_idx);
unlock(&self.mmap);

if let Some((abs, len)) = active_abs_len {
    let result = f(&self.mmap[abs..abs + len]);
    if self.page_generation(active_idx) != gen_before {
        return None;
    }
    return Some(result);
}
```

shared_cache/src/gen_cache.rs
```diff
@@ -0,0 +1,695 @@
...
+        unlock(&self.mmap);
+
+        // Phase 2: lock-free — sealed pages, newest first.
+        for i in 1..self.n_pages {
+            let page_idx = (active_idx + self.n_pages - i) % self.n_pages;
+            let gen_before = self.page_generation(page_idx);
+            if let Some((_, _, slot_idx)) = self.do_probe(page_idx, hash, key) {
+                // Check generation hasn't changed between probe and write.
+                // If it has, rotation zeroed this page already — skip.
+                if self.page_generation(page_idx) == gen_before {
```
jbylund: **Bug (TOCTOU): generation check and TOMBSTONE write are non-atomic — can corrupt a live active-page entry.**

Between `self.page_generation(page_idx) == gen_before` and the `key_hash = TOMBSTONE` write, another worker can acquire the spinlock and run a full `commit_rotation` on `page_idx`: bump its generation, zero its slot table, re-insert survivors, unseal it, advance the counter. When this worker then writes TOMBSTONE into `slot_idx`, it is writing into the **active** page — either corrupting a freshly re-inserted survivor or turning an empty slot into a permanent tombstone that breaks linear-probing chains.

The fix is to do the tombstone write under the spinlock (matching Phase 1), or use a CAS on `key_hash` (expecting the original hash, writing TOMBSTONE) plus a generation re-check inside the lock.

shared_cache/src/gen_cache.rs
```diff
@@ -0,0 +1,695 @@
...
+        let Some((val_off, key_off)) = self.alloc_arena(page_idx, value_bytes.len(), key.len()) else {
+            return false; // arena full
+        };
+        unsafe {
+            let ab = self.arena_base_mut(page_idx);
+            std::ptr::copy_nonoverlapping(value_bytes.as_ptr(), ab.add(val_off as usize), value_bytes.len());
+            std::ptr::copy_nonoverlapping(key.as_ptr(), ab.add(key_off as usize), key.len());
+        }
+        let is_new = existing_idx.is_none();
+        let slot_idx = existing_idx.unwrap_or(idx);
```
jbylund: **Bug (arena accounting): updating an existing key always allocates new arena bytes, orphaning the old ones.**

When `existing_idx` is `Some`, `alloc_arena` still bumps `arena_head`, orphaning the old key+value bytes until the next rotation. `entry_count` tracks the *slot* count (which is stable on updates), so rotation's capacity check fires at the right time — but arena pressure grows on every `set()` call for a hot key regardless of `entry_count`. With the default config (`gen_maxsize ≈ 3333`, arena `gen_maxsize × 8192 ≈ 27 MB`), a single frequently-updated key written ~3300 times in one generation fills the arena, causing `alloc_arena` to return `None` → `do_insert` returns `false` → the new value is silently not stored, stale data continues to be served.

A simple fix: when `existing_idx.is_some()` and the new serialized value fits within the existing `arena_len`, write in-place and skip `alloc_arena`.

shared_cache/src/types.rs
```diff
@@ -0,0 +1,29 @@
...
+    pub result_count: Option<i64>,
+    pub total_cards: Option<i64>,
+}
+
+/// One 64-byte slot in the open-addressing hash table.
+#[repr(C)]
+pub struct RawSlot {
+    pub key_hash: u64,     // EMPTY / TOMBSTONE / normalized xxh3 hash
+    pub expiry_ns: u64,    // Unix epoch nanoseconds; u64::MAX = never expires
+    pub value_hash: u64,   // xxh3(status || body) — fast-path dedup in set()
```
jbylund: **Nit: comment is wrong — `value_hash` is a body-only sampled hash, not `xxh3(status || body)`.**

`content_vh` is computed as `sampled_body_hash(body.as_deref())` — the HTTP status string is never mixed in. The fast-path dedup can therefore silently skip re-inserting a response whose status changed but whose body bytes happened to be identical. In practice this is blocked by the `resp.status.startswith("5")` guard in the Python middleware (5xx responses are never cached), but the comment misleads future readers.

```rust
pub value_hash: u64,   // sampled_body_hash(body) — fast-path dedup in set()
```

api/middlewares/caching_middleware.py
```diff
@@ -132,8 +157,13 @@ def process_response(
...
             return
-        cache_key = self._cache_key(req)
-        if self.cache.get(cache_key) is not None:
+        cache_key = req.context.get("cache_key") or self._cache_key(req)
+        # __contains__ on SharedCache checks only the cuckoo filter (~0.006% FPR), so a false
+        # positive here causes a response to never be cached for that key. A full probe via
+        # .get() would eliminate false positives at the cost of a lock + slot scan per miss.
+        # The same body-hash fast-path used in set() could be applied here to cheaply confirm
+        # before skipping, but the FPR is low enough that the simpler check is worth it.
+        if cache_key in self.cache:
```
jbylund, line 166: **Bug: after `pop()` on a sealed-page entry, this check permanently suppresses re-caching for that key.**

`__contains__` on `SharedCache` only checks the cuckoo filter. `pop()` on a key found only in a sealed page tombstones the slot but — by design — does **not** delete the filter fingerprint (the `pop()` docstring in `gen_cache.rs` confirms this). So after `pop(key)`, `key in self.cache` returns `True` while `get(key)` returns `None`. This line then returns early without re-caching, and the key stays uncacheable until a rotation clears the fingerprint.

This doesn't affect the full `invalidate()` path (which zeros the filter), but any caller using `pop()` for per-key invalidation of sealed-page entries will silently break caching for those keys.

api/middlewares/caching_middleware.py
```diff
@@ -49,27 +58,42 @@ class CachedResponse(NamedTuple):
...
             cache: Optional cache instance. If None, creates an LRUCache with maxsize 10,000.
+                Any object supporting .get(), __contains__, and __setitem__ is accepted.
         """
         if cache is None:
             cache = LRUCache(maxsize=10_000)
-        self.cache: MutableMapping[CacheKey, CachedResponse] = cache
+        self.cache: _CacheProtocol = cache
+        logger.info("CachingMiddleware init pid=%d cache=%s", os.getpid(), type(cache).__name__)
+
+    def invalidate(self: CachingMiddleware) -> None:
```
jbylund, line 73: **TODO gap: `invalidate()` is never called; bulk imports do not flush the HTTP response cache.**

The TODO comment acknowledges this, but the method shipped as a dead stub. `APIResource._clear_caches()` (called after every bulk import) bumps `cache_generation` to invalidate query/engine caches but has no reference to the `CachingMiddleware` instance — it is constructed inside `get_api()` and not exposed. Stale compressed HTTP responses will be served until natural eviction after an import.

Wiring this up requires either passing the middleware instance into `APIResource` at construction, or making the app-level middleware list accessible from within the resource.

jbylund (commented):
Seven findings from a medium-effort review. Four confirmed correctness bugs (two UB paths, one TOCTOU, one arena accounting), plus three lower-severity issues. Detailed inline comments below.

shared_cache/src/gen_cache.rs
```diff
@@ -0,0 +1,790 @@
...
+    }
+
+    /// Returns (arena_offset, arena_len, slot_idx) or None.
+    /// Lock-free safe for sealed pages.
+    fn do_probe(&self, page_idx: usize, hash: u64, key: &[u8]) -> Option<(u32, u32, u32)> {
+        let slot_count = self.slot_count_per_page as u32;
+        let mut idx = (hash as u32) % slot_count;
+        loop {
+            let slot = unsafe { &*self.slot_ptr(page_idx, idx) };
+            match slot.key_hash {
```
jbylund: **Data race: `do_probe` reads `key_hash` non-atomically on sealed pages**

`scan_survivors()` at line 382 explicitly fixes this exact race with `read_key_hash()` (Acquire atomic load) and its comment names the culprit: "`pop()`'s plain write under the spinlock." `do_probe()` has the same exposure — it reads `slot.key_hash` as a plain struct field access on the lock-free sealed-page paths in `get_with()` (lines 571–588), `fast_check()`, and `contains()`, while `pop()` at line 704 writes `TOMBSTONE` to sealed-page slots under the spinlock with no atomics.

This is UB under Rust's memory model. On x86/ARM64 aligned 8-byte reads are naturally atomic at the hardware level so the practical risk is low, but LLVM is entitled to optimize across non-atomic reads in ways that break the assumption.

Fix: replace `slot.key_hash` here with `read_key_hash(self.slot_ptr(page_idx, idx) as *const u8)`. The atomic load is a free operation on x86 (same instruction as a plain load).

api/middlewares/caching_middleware.py
```diff
@@ -49,27 +58,42 @@ class CachedResponse(NamedTuple):
...
+        """Clear all cached entries, delegating to the inner cache's own method."""
+        # Not yet wired into APIResource._clear_caches() — bulk imports do not currently
```
jbylund, line 75: **`invalidate()` not wired to bulk imports; no TTL means stale responses are permanent**

The three bulk import paths in `api_resource.py` (lines 946, 1766, 2290) call `_clear_caches()`, which bumps `_cache_generation` to invalidate query caches but has no handle to `CachingMiddleware`. The HTTP response cache is never flushed on import.

The severity is higher than the comment suggests: `SharedCache` is opened without `default_ttl_secs`, so all entries have `expiry_ns = u64::MAX` (never expire). A stale search-result page cached before a bulk import will be served until that specific slot is displaced by the 10,000-entry LRU pressure. On a low-traffic instance with a warm, diverse cache, that could be indefinitely.

The fix is straightforward — pass the `CachingMiddleware` instance into `APIResource` at construction time and call `self.caching_middleware.invalidate()` from `_clear_caches()`. The method and its delegation to `SharedCache.invalidate()` are already implemented; only the wiring is missing.

jbylund: I'm not doing this now.

PR #580, opened 2026-07-02
Use Cache Control Header to Control Caching Behavior in Local Cache

jbylund:
## Summary
- Replace the hardcoded `_UNCACHED_PATHS` path list in `CachingMiddleware` with a general `Cache-Control: no-store` convention — the middleware now skips storing (and skips checking) any response that sets that header, rather than special-casing `/random_search` by path.
- Add a `set_no_store_header()` helper in `api_resource.py` and apply it to `get_pid` and `random_search`, each of which now accepts an optional `falcon_response` kwarg.
- Remove the dead `db_ready()` endpoint and the bogus `_handle()` early return for `("db_ready", "pid")`. Neither string was ever a real route (the real pid endpoint is `/get_pid`), and nothing ever called `/db_ready` over HTTP — `db_ready()` was only ever invoked directly from tests. Deleted the method and its 3 tests rather than wiring it up as a real endpoint.

## Why
Per-path exclusion lists don't scale and are easy to forget when adding new non-cacheable endpoints. Marking non-cacheable responses via the standard `Cache-Control` header keeps the caching decision local to the handler that knows it shouldn't be cached.

While reviewing, found that `db_ready` was unreachable via HTTP due to a pre-existing bug in `_handle()`, making it dead code. Since nothing depends on it, removed it rather than fixing it.

## Test plan
- [x] `test_caching_middleware.py` updated: `test_cache_bypass_sets_no_header` split so the removed uncached-path case is dropped and a disabled-cache case remains.
- [x] New `test_no_store_response_not_cached` verifies responses with `Cache-Control: no-store` are not written to the cache.
- [x] Removed `db_ready` tests in `test_api_resource.py` and `test_integration_testcontainers.py`.
- [x] Full unit suite (465 tests) passes.

### DIFF
--- a/api/api_resource.py
+++ b/api/api_resource.py
@@ -702,9 +702,6 @@ class APIResource:
 
         path = req.path.strip("/") or "_root"
 
-        if path in ("db_ready", "pid"):
-            return
-
         logger.info(
             "Handling request for %s / |%s| / response id: %d",
             req.relative_uri,
@@ -856,21 +853,6 @@ class APIResource:
         """
         set_no_store_header(falcon_response)
         return os.getpid()
-
-    def db_ready(self, *, falcon_response: falcon.Response | None = None, **_: object) -> bool:
-        """Return true if the db is ready.
-
-        Returns:
-        -------
-            bool: True if the database is ready, False otherwise.
-
-        """
-        set_no_store_header(falcon_response)
-        records = self._run_query(
-            query="SELECT relname FROM pg_stat_user_tables",
-        )["result"]
-        existing_tables = {r["relname"] for r in records}
-        return "migrations" in existing_tables
 
     def setup_schema(self, *_: object, **__: object) -> None:
         """Set up the database schema and apply migrations as needed."""
--- a/api/api_resource.py
+++ b/api/api_resource.py
@@ -446,6 +446,13 @@ def set_cache_header(falcon_response: falcon.Response | None, duration: timedelt
     falcon_response.set_header("Cache-Control", f"public, max-age={seconds}")
 
 
+def set_no_store_header(falcon_response: falcon.Response | None) -> None:
+    """Set Cache-Control: no-store on a Falcon response to prevent CDN and browser caching."""
+    if falcon_response is None:
+        return
+    falcon_response.set_header("Cache-Control", "no-store")
+
+
 @cached(cache=LRUCache(maxsize=16))
 def _build_base_html(critical_css: str, site_name: str) -> str:
     """Read index.html and inject critical CSS and site name. Cached per (critical_css, site_name) pair."""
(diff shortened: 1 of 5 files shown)


api/api_resource.py
```diff
@@ -839,24 +846,26 @@ def maybe_json_dump(v: object) -> object:
...
-    def db_ready(self, **_: object) -> bool:
+    def db_ready(self, *, falcon_response: falcon.Response | None = None, **_: object) -> bool:
         """Return true if the db is ready.
 
         Returns:
         -------
             bool: True if the database is ready, False otherwise.
 
         """
+        set_no_store_header(falcon_response)
```
jbylund: This call never fires for real HTTP requests. `_handle()` (line ~531) has a pre-existing early return — `if path in ("db_ready", "pid"): return` — that fires *before* the dispatch block reaches `db_ready(falcon_response=resp, ...)`. So a live `GET /db_ready` never invokes this method at all, and `Cache-Control: no-store` never gets set on that endpoint's response — the exact thing this PR is meant to fix for `db_ready`. (`get_pid` is unaffected: its real route is `/get_pid`, used as the Docker healthcheck in `docker-compose.yml:96`, which doesn't match the `"pid"` string in that check.) All existing `db_ready()` tests call it directly rather than through `_handle`, so this gap has no coverage.

jbylund (commented):
Automated review pass — 3 findings inline. Nothing blocking, but #1 undermines this PR's stated goal for `db_ready`.

api/middlewares/caching_middleware.py
```diff
@@ -110,8 +105,6 @@ def process_request(self: CachingMiddleware, req: falcon.Request, resp: falcon.R
         """
         if not settings.enable_cache:
             return
-        if self._is_uncached(req):
```
jbylund, line 113: Removing this check changes `/random_search` from "structurally impossible to serve from cache" to "safe only as long as `set_no_store_header` always runs before the response is finalized." `_cache_key` has no per-request nonce, so if a stale entry ever exists under this key — e.g. written by an old worker during a rolling/blue-green deploy, before this PR's no-store header existed — `/random_search` would start serving that cached entry as a repeated, non-random hit until it's evicted. Worth confirming this repo's deploy process can't produce that window (or adding a regression test that drives a `/random_search` request through `process_request` end-to-end, since the deleted parametrized test case for this exact behavior wasn't replaced).

api/api_resource.py
```diff
@@ -446,6 +446,13 @@ def set_cache_header(falcon_response: falcon.Response | None, duration: timedelt
     falcon_response.set_header("Cache-Control", f"public, max-age={seconds}")
 
 
+def set_no_store_header(falcon_response: falcon.Response | None) -> None:
```
jbylund, line 449: Minor: this duplicates `set_cache_header` right above it (same `if falcon_response is None: return` + `set_header("Cache-Control", ...)` shape) rather than reusing it. Not blocking, but if the null-check or header-setting logic ever changes, both functions need updating in sync.

jbylund: I'm fine with that level of duplication.

PR #794, opened 2026-07-28
Refactor: Mount Data-Management Handlers as a Child Resource

jbylund:
Step 3 of the plan in `docs/issues/local-api-resource-routing-redesign.md`, plus four extractions
taken along the way. **`api/api_resource.py` goes from 2,586 to 1,306 lines.**

Commits are independently reviewable and land in a deliberate order — the extractions first, so the
admin move lands against a smaller file.

## The problem

Registration had no way to express "this handler is not part of the public API." Every public method
became a route in one flat namespace, and the only lever for hiding one was a leading underscore,
which also lies about a method's Python visibility — `setup_schema` is called from `__init__` and from
tests and could not be renamed to hide it.

`api/admin_resource.py` takes the eleven data-management handlers, the eleven helpers only they reach,
`prefer_score_tuner`, and the module-level constants and imports no other part of the file used.
`APIResource` mounts it under a path prefix using the *same* `build_route_table` its own registration
uses, so a child cannot end up registered by different rules than its parent.

## The seam, derived rather than guessed

Computed from the AST, then asserted in tests: **five methods and four handles**.

| crossing | |
| --- | --- |
| methods | `_reload_engine`, `_run_query`, `_serve_static_file`, `_set_statement_timeout`, `_setup_complete` |
| handles | `_conn_pool`, `_cache_generation`, `_last_import_time`, `_setup_complete_cache` |

The child holds a parent reference for those. **That is a routing boundary, not a decoupling**, and
the module docstring says so — pretending otherwise would mean an `AppContext` refactor this does not
need.

Two refinements over the original analysis: `_serve_static_file` is a crossing only because
`prefer_score_tuner` moved, and `_setup_complete_cache` is one nobody had listed — admin-only
`_clear_caches` invalidates a cache the public `_setup_complete` owns.

Also verified: every module-level helper the moved code needs is admin-only, so the import runs one
way and there is no cycle. That was the structural risk.

## The listing is the part that could have gone wrong quietly

`advertise=False` is passed **once at the mount**, not repeated on each handler, so forgetting it is a
property of the single call site rather than a hole in one route. `build_routes_listing` filters on
the flag. Without that the mount would be a URL prefix rather than a boundary, and every other test
would still pass.

```
listing entries: 30 -> 17     admin entries in the listing: none
GET /_admin/nope   and   GET /totally/bogus   ->   byte-identical 404
```

## Four extractions taken along the way

| module | what moved |
| --- | --- |
| `api/utils/routing.py` | `max_positional_args`, the registration loop, the 404 listing — a child needs all three |
| `api/utils/site_name.py` | Host-header display names: TLD set, validation patterns, dictionary word-splitting |
| `api/utils/page_rendering.py` | template paths, HTML fragments, asset content-hashes, minification, both page builders |
| `api/utils/caching.py` | the settings-aware `cached` decorator, which three destinations use |

I looked at pulling the remaining module-level constants into a `constants.py` and decided against
it: the six left are used only by the class in that same file, so it would add an import for no
cohesion gain.

## Tests

2,189 pass. `api/tests/test_admin_mount.py` adds 24, including **the public route table as a literal
set** — the regression guard for the class of bug this replaces. Eight bare `hasattr` smoke checks are
deleted rather than ported; the literal supersedes them. Site-name and page-rendering tests moved to
sit with their modules.

Two `test_engine_unit.py::TestExplain` failures are pre-existing — confirmed identical on `main` in a
throwaway worktree, and a stale locally-built `card_engine`, not something in this diff.

## Notes

- `get_migrations` loses its route; `db_utils.get_migrations()` stays for in-process use. Nothing
  called it over HTTP.
- Ruff wanted `falcon` moved under `TYPE_CHECKING` in the new module. That would break registration,
  since handlers annotate `falcon_response: falcon.Response | None` and those resolve at mount — the
  trap `routing.py` documents. Kept at runtime with the same `noqa` pattern `api_resource` uses.
- `CLAUDE.md`'s architecture section described a file that no longer exists in that shape; updated.


jbylund:
Superseded by #963, which was verified to be a byte-identical diff to this PR's stack (#794 → #961 → #962) recreated as a single PR against `main`, plus follow-up review fixes. Closing without merging since #963 already carries these changes.

PR #797, opened 2026-08-01
Report Acquire Facts and Executor Counters From explain, and Measure Cost-Model Agreement

jbylund:
## Summary

Behavior-neutral instrumentation: `explain` reports the feature vector `cost::plan_cost` actually
consumes, and `explain_analyze` reports per-plan executor counters and phase timings. Plus
`bench_cost_model_agreement.py`, the first consumer.

Nothing about routing, costing or results changes. This is visibility — and it already found a
selected plan that is priced at 795 ns, spends 9.4 µs producing nothing, and hands the query back.

## Why this first

The cost model can only be improved against measurements, and today there is no way to see inside a
query from outside the engine. Two consequences, both hit while doing the follow-on work:

- **A cost error is invisible until it becomes a latency error.** A plan whose arm is 2x wrong is
  picked correctly right up until something competes closely, and then it is a mystery. With the
  feature vector exposed, the arm can be checked directly against the counter that realizes it.
- **Cross-version comparison is a black box.** Benchmarking a branch against `main` currently yields
  total wall time and nothing else, so a 2% regression cannot be attributed to acquire, prepare, the
  match loop, or emit. Landing the instrumentation on `main` is what makes every later change
  measurable, which is why it goes first rather than riding along with a behavior change.

## What is exposed

`explain`'s `acquire` gains the model's own inputs — `eval_domain`, `scan_units`, `matches`,
`n_cards`/`n_printings`, `residual_tier_ns100`, the compose build terms
(`broadcast_printings`/`scatter_printings`/`project_printings`/`popcount_words`), `compose_paging`,
and the acquire branch (`count_source`). A calibration fit can then regress on exactly the vector the
engine reads instead of a reconstruction of it.

`explain_analyze`'s per-plan entries gain `cards_visited`, `printings_scanned`, `matches_pushed`, the
phase split `ns_prepare` / `ns_setup` / `ns_loop` / `ns_finish` against `ns_round_total`, and the two
ground-truth fields `result_total` and `paging_taken`, plus `materialize_ns` and `routed_ns`. The
counters are what let a FEATURE be checked rather than inferred: `matches` is supposed to predict
`matches_pushed`, `scan_units` to predict `printings_scanned`. Where they disagree, no coefficient
can repair it.

A plan that entered a fastpath and turned back reports `declined_ns` instead of `trials_ns` — raw
per-round wall time, kept in its own field because every consumer reduces `trials_ns` with `min` as
"what this plan costs to run", and a decline is not a run. Its counters and phase timings stay zero
(no executor ran) but `paging_taken` names the gate, which is the only thing that distinguishes a
cheap turn-back from an expensive one. `PhysicalPlan::applicable` is at least as strong as the guard
`run_query_with_plan` re-checks, so `None` from a ranked plan is always a runtime decline and never
a missed applicability test — that is what makes the field well-defined.

`ns_setup` is split from the match loop rather than folded into it because it is neither small nor
modelled: `run_query_streamed` zeroes an `n_cards`-long counts buffer (~126 kB on the real corpus)
regardless of how few candidates it is about to visit. Under a `printing_compose` acquire that is 0%
of the run at the median and **65% at p90** — the selective case, previously landing in the
unaccounted remainder. It is also the one phase with an obvious cost term waiting for it, since the
dominant part scales with `n_cards`, which `PlanFeatures` already carries. The phases are contiguous,
so four `Instant::now()` calls bound all three: no clock reads were added to split it out.

## The response shape, before and after

Both blocks are real output, same query (`t:creature`, `unique=card`, `orderby=edhrec`,
`num_warmups=1, num_trials=3`) against the real corpus — the "before" produced by checking out
`origin/main`'s `card_engine/src` and rebuilding the extension, not hand-written.

**Before** — a bare list, three fields per plan:

```json
[
  { "plan": "PlanePopcountOrder", "predicted_ns": 12048.362500000001,
    "trials_ns": [511917, 513208, 499000] },
  { "plan": "StreamedSelect",     "predicted_ns": 54182.7,
    "trials_ns": [1118750, 1147667, 1145000] },
  { "plan": "GatheredScan",       "predicted_ns": 112805.5,
    "trials_ns": [2556791, 2521959, 2670292] }
]
```

You can see the ranking and you can see the wall time. You cannot see *why* either number is what it
is, and there is nowhere to put a query-level fact, which is why the top level had to become a dict.

**After** — `{"acquire": ..., "plans": [...]}`:

```json
{
  "acquire": {
    "count_source": "plane",
    "narrowed_repr": "none",
    "acquire_ns": [17000, 15625, 16709],
    "routed_ns": [465000, 453917, 456916],
    "eval_domain": 17317,
    "n_cards": 31508,
    "matches": 17317,
    "n_printings": 97206,
    "scan_units": 17317,
    "residual_tier_ns100": 0,
    "limit": 50,
    "offset": 0,
    "broadcast_printings": 0,
    "scatter_printings": 0,
    "project_printings": 0,
    "popcount_words": 0,
    "compose_paging": "Gather"
  },
  "plans": [
    {
      "plan": "PlanePopcountOrder", "predicted_ns": 12048.362500000001,
      "materialize_ns": 0.0, "picked": true,
      "trials_ns": [456000, 493625, 472166],
      "cards_visited": 0, "printings_scanned": 0, "matches_pushed": 0,
      "ns_setup": 0, "ns_loop": 0, "ns_finish": 0, "ns_prepare": 0,
      "ns_round_total": 456000, "result_total": 17317, "paging_taken": ""
    },
    {
      "plan": "StreamedSelect", "predicted_ns": 54182.7,
      "materialize_ns": 85862.15000000001, "picked": false,
      "trials_ns": [1157417, 1121625, 1109417],
      "cards_visited": 17317, "printings_scanned": 45976, "matches_pushed": 17317,
      "ns_setup": 95666, "ns_loop": 896667, "ns_finish": 11250, "ns_prepare": 105208,
      "ns_round_total": 1109417, "result_total": 17317, "paging_taken": ""
    },
    {
      "plan": "GatheredScan", "predicted_ns": 112805.5,
      "materialize_ns": 85862.15000000001, "picked": false,
      "trials_ns": [2504875, 2319625, 2466416],
      "cards_visited": 17317, "printings_scanned": 45976, "matches_pushed": 17317,
      "ns_setup": 334, "ns_loop": 2196958, "ns_finish": 16042, "ns_prepare": 105292,
      "ns_round_total": 2319625, "result_total": 17317, "paging_taken": ""
    }
  ]
}
```

`explain` returns the same two keys; its `plans` entries stop at `predicted_ns`/`materialize_ns`/
`picked`, and its `acquire.routed_ns` is empty because it runs nothing.

Both blocks predate `declined_ns` and are left exactly as captured rather than re-measured: they are
a matched pair taken on one machine in one state, and replacing only the "after" half would turn a
shape comparison into a fabricated speedup. No plan declines on this query, so the field would read
`[]` throughout anyway — it is shown with real content under "What it already found" below.

Four things this one response makes checkable that the old one did not:

- **`predicted_ns` is byte-identical across the two builds** (`12048.362500000001`, `54182.7`,
  `112805.5`). Costing did not move.
- **The counters agree across executors.** `StreamedSelect` and `GatheredScan` both report
  `17317 / 45976 / 17317`. They are one counter under one name and `scan_units` has one definition,
  so a disagreement would mean at most one is the valid comparison —
  `materializing_plans_agree_on_the_counters_they_share` is what pins it.
- **A feature can be checked against what happened.** `matches: 17317` predicts
  `matches_pushed: 17317` exactly here; `scan_units: 17317` against `printings_scanned: 45976` is a
  2.65x under-count, visible as itself rather than as a coefficient that will not settle.
- **The winner is 38x cheaper than predicted, and the phase split says where.** `PlanePopcountOrder`
  predicts 12 µs and runs in 456 µs. It is a fast path, so its counters are zeros *by definition* —
  and `ns_round_total: 456000` with `ns_loop: 0` is exactly the "uninstrumented, not idle" signal
  documented on `PlanTrial::phases`. Closing that is
  `docs/issues/local-engine-instrument-fast-paths.md`.

Note `ns_prepare` ≈ 105 µs against a `StreamedSelect` round of 1109 µs — ~9.5% of the run, in a term
`plan_cost` does not carry at all. That is the gap this instrumentation exists to size.

## What it already found

`paging_taken` exists because `compose_paging` is a prediction — `acquire_plan_features` re-derives
the permutation lookup and the orderby-walk test that `printing_compose_fastpath` then decides again
at run time, and nothing checked the two agree. They now can be. Over 120s of the real corpus
(31,858 queries), split by acquire branch:

    acquire              strategy runs    disagreements    walk-declined
    printing_compose             4,398             0    0%              1
    card_range_popcount            635           635  100%              0
    printing_range_scan            958           958  100%              3

`walk-declined` is broken out rather than pooled into either column — it is the one cell where the
prediction is allowed to be inexact, so counting it as agreement would hide it and counting it as
drift would misname it.

**The two range rows are the defect.** Both cost a *competing* `PrintingCompose` with the untouched
`mk_plan_feats` default of `Gather`, so they price an arm it never runs — `Gather` is
O(`eval_domain`), `Perm` is O(`page_span`/selectivity). The fix is to hand those branches a result
total, not to reconcile two predictions. Not done here: this PR reports.

**The single `printing_compose` walk-declined row is not drift**, and tracing it is what produced
`GatherWalkDeclined`. The two *availability* tests are byte-identical by construction, so they cannot
disagree; what acquire cannot see is one step further on, because it never composes and so cannot
know whether the walk it predicts will *succeed*. The walk declines on the null-value tail or a page
past the value structure and falls into the gather. So `compose_paging` is an **upper bound on
`OrderbyWalk`** and exact in every other cell — one-directional, structural. That gather now reports
`GatherWalkDeclined` rather than a bare `Gather` indistinguishable from a real misprediction, and the
harness reports its rate on its own line: those runs pay a walk attempt *and* a gather while priced
as a walk, which is a real (small) under-costing worth its own follow-up.

This supersedes b83769e's commit-message table, which put the compose-acquired branch at 20% wrong.
That 20% was entirely `predicted Gather -> actual Decline` rows. Pooling the three acquire branches
into one number was hiding a cell that is wrong by construction inside one that is right.

### The declines are 21% of compose entries, and one gate is expensive

The table above counts only composes that reached a paging strategy — which is every compose the
harness could see, because `explain_analyze` dropped the stats of any plan whose round produced no
page. So the four decline gates were recorded in Rust, asserted in-crate, and discarded before
anything outside the crate could count them: a census of *successful* composes presented as a census
of composes. `declined_ns` closes that, and the missing fifth of the picture is not uniform:

    plan               gate                       n   median µs   p90 µs
    PrintingCompose    DeclineSparseExact       959        12.2     15.2
    PrintingRangeScan  (no gate label)          679         0.0      0.0
    PrintingCompose    DeclineBroad             397         0.2      0.3
    PrintingCompose    DeclineSparseEstimate    236         0.0      0.1

**The blank gate column in row 2 is a defect this PR then fixed, not a current finding.** That table
was captured before `printing_range_fastpath` labelled its exits; at the time only the compose
fastpath named a gate, so a declining `PrintingRangeScan` arrived as `NotEntered` — indistinguishable
from never having been tried — and rendered as an empty cell. Every exit of both fastpaths now
records itself, so those 679 rounds resolve into `RangeSelective`/`RangeSparse`/`RangeUnalignedPrice`/
`RangeNoPermutation`. The table is left as captured rather than re-measured for the same reason the
before/after JSON above is: it is one machine in one state, and refreshing half of it would make the
comparison worse, not better. The counts and costs in the other three rows are unaffected — the
labelling change adds no work to any path.

Two of those exits — `RangeNotBare` and `RangePermutationStale` — are tripwires rather than expected
outcomes: the first contradicts `PrintingRangeScan.applicable`, which already requires bare range
bounds, and the second means a permutation built against a different store. Compose's
`NotComposable` is the same class, since `PrintingCompose.applicable` *is* the structural test the
fastpath re-checks. All three sit in `bench_cost_model_agreement.py`'s `IMPOSSIBLE_GATES`, so a
non-zero count prints as a `DEFECT:` line rather than as three more cheap rows. None of them fires
today.

Three of the four compose gates gate *before* the compose and cost nothing, which is the design
working. `DeclineSparseExact` gates *after*, off the real total rather than the estimator's bound, so
it has already composed the printing bitmap and throws it away. The pre-compose sparse gate catches
236 while **959 get past it** to pay a full compose and discard it — a 4:1 miss rate on a gate whose
entire job is catching those.

This is not confined to plans the router rejected. A real response, `cn:312` at `unique=card`:

```json
{
  "plan": "PrintingCompose", "predicted_ns": 795.54, "picked": true,
  "trials_ns": [], "declined_ns": [9541, 9375, 9417],
  "cards_visited": 0, "ns_loop": 0, "ns_round_total": 0,
  "result_total": 0, "paging_taken": "DeclineSparseExact"
}
```

The router **picked** this plan, priced it at 795 ns, and it spent 9.4 µs producing nothing before
the general path ran the query anyway — a 12x under-cost on a selected plan, paid on every such
query. That is precisely the "cost error invisible until it becomes a latency error" this PR opens
with, and it was invisible until `declined_ns` existed.

The fix is estimator tightening rather than a new cost term — the gate exists and fires, it just
fires too late — so it is its own follow-up. Not done here: this PR reports.

## Measurement discipline

`explain_analyze` runs `n + 2` participants per round — every applicable plan, plus the acquire step
and the routed path — in a **seeded-shuffled order**, rather than rotating the plans and pinning the
other two ahead of them.

Pinning them was biased, and specifically so: each pre-warms a *subset* of the plans it is meant to be
compared against. `run_query_routed` dispatches into the picked plan's executor, so a fixed slot ahead
of the plan loop pre-warmed exactly the plan whose selection `picked` exists to question. And on a
candidate acquire, `acquire_plan_features` calls the very `prepare_candidates` that `StreamedSelect`
and `GatheredScan` are about to call and the four fast paths never do — so the bias was present for
one acquire class and absent for another, not even washing out as a constant offset.

Shuffled rather than rotated because a rotation only moves the cut point: cyclic adjacency is fixed,
so every participant keeps the same immediate predecessor round after round. Seeded from a constant,
so ordering stays reproducible for an A/B against another build.

What the shuffle gives up, stated because the trade is not free in both directions: rotation was
*position-balanced* — over `n` rounds every participant sat in every slot exactly once — and an
independent per-round shuffle is not. At `num_trials = 3` a participant can draw the warm tail of the
round twice by chance, and since every consumer reduces `trials_ns` with `min`, that lower sample is
the one that survives. The trade is still right — the bias removed was directional and unequal across
acquire classes, this one is zero-mean — but it shrinks only with rounds, so `num_trials` is now
load-bearing. `bench_cost_model_agreement.py` uses 7; a 2-3 trial run is not a fair head-to-head
between plans within ~10% of each other.

**"Zero-mean and shrinks with rounds" is true within one query and not across a sweep**, which is
[#801](https://github.com/jbylund/sylvan_librarian/issues/801)
([doc](docs/issues/00801-engine-explain-analyze-shuffle-seed.md)), found in review of this PR and
deliberately left out of it. The RNG is constructed inside `explain_analyze` from a constant, so
every call with the same participant count draws the identical sequence of permutations. The
resulting bias attaches to rank position, which correlates with plan identity, so it survives into a
table keyed by plan and *does not* average out over more queries — it is common-mode across samples
rather than noise within them. That applies to every table in this PR produced by
`bench_cost_model_agreement.py`: the paging-agreement counts and the decline census are ordering-
sensitive in their timings, though not in their counts, which is what the findings above turn on.
The fix is to mix a per-query value into the seed; it is a methodology change to the harness, and
landing the instrumentation first is what makes it measurable at all.

**Caveat for anyone holding earlier numbers from this branch: `trials_ns` collected before the shuffle
is not comparable to `trials_ns` collected after it.**

## Correctness

Instrumentation must change nothing observable.

- **648 query shapes** (18 predicates x 3 distinct-ons x 4 orderbys x 3 pages), branch vs `main`:
  **0 total mismatches, 0 first-page mismatches.**
- `cargo test`: 145 pass debug, 144 release. 2,228 Python unit tests, 159 engine unit tests.
- `paging_taken` lives in its own `PAGING_TAKEN` thread-local rather than sharing a slot with the
  counters. It is the one field written by something that is not an executor, and sharing forced both
  publishers to write 8 of 10 fields and inherit the rest through `..c.get()` — making every
  publisher responsible for not wiping a field it does not own, which is the arrangement the 49-of-600
  leak came out of. Both executors now write `PhaseStats` whole and can inherit nothing;
  `take_phase_stats` reassembles the two and consumers see no seam. `note_paging_taken` also drops
  from a read-modify-write of the whole ~80-byte struct to a one-byte store, which matters because it
  runs on the production compose path.
- `PENDING_PREPARE_NS` is a value in flight between two calls rather than a stat, so nothing clears
  it — the receiving executor consumes it with `replace(0)`. That invariant spans three functions and
  the failure is silent (an unconsumed value is read by the *next* participant and reported as its
  `ns_prepare`), so `timed_prepare_candidates` asserts it in debug.
- The diagnostic labels (`paging_taken`, `count_source`, `narrowed_repr`) are enums with explicit
  `label()` methods, verified byte-identical against the built extension — the Python surface did not
  move, including `PagingTaken::NotEntered -> ""`. Two "the label is one of these five" assertions
  became tautologies and were deleted.
- Counters are published through a single helper because the walk has several early returns, and each
  one has to leave the stats behind or the phase accounting silently attributes work to nothing.
- `ns_setup + ns_loop + ns_finish` is asserted as a sum bounded by the round, not as an equality:
  `run_query_with_plan`'s dispatch and the applicability checks sit outside all three, so the
  shortfall is a residual to size rather than an invariant that should reach zero.

### Regression tests, each verified against the failure it exists to catch

- `materializing_plans_agree_on_the_counters_they_share` — the two executors publish into one
  `PhaseStats` under one set of names, and `scan_units` has one definition, so if they count
  differently at most one is the valid comparison. Reverting the counter-placement fix fails at
  `streamed 15133 vs gathered 1254`. Its first draft passed vacuously (plane/index specs narrow
  exactly, so `card_pass` never rejected anything); inexact narrowings fixed that, and "the continue
  fired somewhere" is now itself asserted. `matches_pushed` is additionally anchored to the total each
  run actually returned, in both executors — plan-vs-plan alone cannot catch the two being wrong the
  same way, and in artwork mode the counter sits downstream of the group dedup, so a counter moved
  above it would still agree plan-to-plan while both disagreed with reality.
- `plan_stats_never_leak_between_participants` — driven through `explain_analyze`, because the
  invariant lives in *its* loop. Two leaks, and dropping either half of `take_phase_stats` reproduces
  the corresponding one: an uninstrumented plan reads its predecessor's `PHASE_STATS`, and — the leak
  actually observed at 49 of 600 — `paging_taken` crosses from the plan that recorded it to a later
  one. The second is structural: only `printing_compose_fastpath` writes that label, and it must
  survive a materializing publish *within* one run, since the routed path's "compose declines,
  records its gate, falls through to a materializing plan" sequence depends on it. Its lifetime is
  therefore deliberately wider than any single publisher, and the per-participant clear is the only
  thing bounding it to one run instead of "this thread, ever".
- `declining_plans_report_their_gate_through_explain_analyze` — a decline must reach the wire, not
  just the thread-local. Asserts empty `trials_ns` against non-empty `declined_ns`, zero counters and
  zero timings (no executor ran, so anything else is a leak), and a compose label that names a gate
  rather than the `NotEntered` default meaning "never tried". Its sweep reaches both
  `DeclineSparseEstimate` and `DeclineSparseExact`; before `declined_ns` every row it checks was
  discarded by `explain_analyze` before a caller could see it.
- `compose_paging_prediction_matches_the_branch_taken` now asserts per-arm legal labels, which is
  *stricter* than the equality it replaced: a bare `Gather` under a predicted `OrderbyWalk` is now a
  failure. Its sweep reached the walk-declined cell zero times until a narrow composable spec was
  added — only 2 of its 7 specs are printing-composable at all, now noted in the list.

## Performance

No measurable difference, and the measurement says so rather than the intent.

    order                          origin/main    branch
    branch first                       197.9 ms   185.7 ms   branch -6.2%
    origin/main first                  199.2 ms   205.8 ms   branch +3.3%

The two orderings disagree in sign — whichever build runs first is faster — so ordering drift of
about +/-5% dominates and no real difference is resolvable at this scale. Reported both ways rather
than picking the flattering one.

The production-path cost is ~5 clock reads and a thread-local write per query on the two materializing
plans. Deliberately not feature-gated: `explain_analyze` is a `#[pyo3]` method on the shipped
extension, so gating the instrumentation out of release builds would return zeros in exactly the build
worth diagnosing.

## Also included

`bench_cost_model_agreement.py` reports measured/predicted per (plan, acquire branch), sampling
`unique` evenly across card/printing/artwork rather than the 75/20/5 of real traffic — distinct-on
changes which plans are applicable and what `scan_units` means, so under-sampling two thirds of it
hides exactly the cells most likely to be wrong. Its `prepare_candidates` and setup shares cover only
the two materializing plans, because no other plan has those phases at all — which is not the same as
spending 0% of its run in them, and pooling the two says the wrong thing.

It also names the cells it suppresses. Dropping `n < 8` cells silently, and then reporting
`passing/total` over the survivors as if it were coverage, is the same under-sampling bias the script
argues against applied to its own report — so suppressed cells are listed with their sample counts
("n=6 of 8" and "n=1" call for different responses) and the ratio says "reported cells". Measured
honestly: at the default budget nothing is suppressed at all (31,858 samples, smallest cell n=359),
so this is a guard against short runs rather than a live defect.

`bench_candidate_materialize.rs` and `docs/issues/local-engine-candidate-materialize.md` cover the
materialization term `explain` now reports (bitmap beats sort from ~64 candidates; **do not build the
k-way merge** — it loses at every size).

`docs/issues/local-engine-instrument-fast-paths.md` writes up the gap this PR does not close: the four
fast paths report features nothing checks. It needs three different counter vocabularies, not four
missing increments, and one prerequisite — per-acquire-branch feature semantics. The cheap first step
is `run_query_streamed_popcount`, one site covering two plans against a `popcount_words` feature that
already exists.

`docs/issues/00801-engine-explain-analyze-shuffle-seed.md` records the shuffle-seed bias described
under *Measurement discipline* — the one review finding from this PR held back rather than fixed in
it, since it changes how the harness measures rather than what the engine reports.

### DIFF
--- a/api/tests/test_engine_unit.py
+++ b/api/tests/test_engine_unit.py
@@ -1348,12 +1348,12 @@ class TestExplain:
         # Card mode returns one row per matching card, so the pushes ARE the distinct-card total.
         assert gathered["matches_pushed"] <= acquire["n_cards"]
         # The phase split must account for the round, or a phase's cost is attributed to nothing.
-        # The three phases are disjoint sub-intervals of the round, so they can only sum to less
-        # than it -- the shortfall is the unmodelled remainder this instrumentation exists to size.
+        # The phases are disjoint sub-intervals of the round, so they can only sum to less than it --
+        # the shortfall is the unmodelled remainder this instrumentation exists to size.
         # NOT asserted individually as > 0: they are timings, and on a corpus this small
         # `prepare_candidates` really does complete inside the clock's granularity and report 0.
         assert gathered["ns_round_total"] >= gathered["ns_loop"]
-        phases = gathered["ns_prepare"] + gathered["ns_loop"] + gathered["ns_finish"]
+        phases = gathered["ns_prepare"] + gathered["ns_setup"] + gathered["ns_loop"] + gathered["ns_finish"]
         assert phases <= gathered["ns_round_total"], "phase split exceeds the round it came from"
 
         # `result_total` is ground truth for the analyzed run itself, which is the whole reason it
--- a/api/tests/test_engine_unit.py
+++ b/api/tests/test_engine_unit.py
@@ -1315,6 +1315,14 @@ class TestExplain:
             # num_trials recorded.
             assert row["trials_ns"] == [] or len(row["trials_ns"]) == 3
             assert all(t >= 0 for t in row["trials_ns"])
+            # A decline is reported as its own thing rather than as an absence: it produced no page,
+            # so it is not a trial, but it still cost wall time and named the gate that fired. The
+            # two are mutually exclusive because a decline is deterministic for one query and store.
+            assert not (row["trials_ns"] and row["declined_ns"]), f"{row['plan']} reported both trials and declines"
+            if row["declined_ns"]:
+                assert len(row["declined_ns"]) == 3
+                # Nothing executed, so a non-zero counter here would be another plan's.
+                assert (row["cards_visited"], row["matches_pushed"], row["ns_round_total"]) == (0, 0, 0)
         assert any(row["plan"] == "GatheredScan" and len(row["trials_ns"]) == 3 for row in analyzed["plans"])
 
     def test_explain_analyze_counters_track_the_features_that_predict_them(self, engine: QueryEngine) -> None:
--- a/api/tests/test_engine_unit.py
+++ b/api/tests/test_engine_unit.py
@@ -1258,13 +1258,31 @@ class TestExplain:
 
     def test_explain_ranks_applicable_plans_ascending(self, engine: QueryEngine) -> None:
         filters = parse_scryfall_query("t:creature")
-        rows = engine.explain(filters=filters, unique="card", orderby="edhrec", direction="asc", limit=50, offset=0)
+        result = engine.explain(filters=filters, unique="card", orderby="edhrec", direction="asc", limit=50, offset=0)
+        rows = result["plans"]
 
         assert rows, "GatheredScan is always applicable"
         assert {"plan", "predicted_ns"} <= rows[0].keys()
         predicted_ns = [row["predicted_ns"] for row in rows]
         assert predicted_ns == sorted(predicted_ns)
         assert any(row["plan"] == "GatheredScan" for row in rows)
+
+    def test_explain_reports_the_features_the_cost_model_consumes(self, engine: QueryEngine) -> None:
+        """The acquire vector is the point of the diagnostic.
+
+        A calibration fit regresses on exactly what `cost::plan_cost` reads, rather than on a
+        reconstruction of it that can drift.
+        """
+        filters = parse_scryfall_query("t:creature")
+        acquire = engine.explain(filters=filters, unique="card", orderby="edhrec", direction="asc", limit=50, offset=0)["acquire"]
+
+        assert acquire["count_source"], "every query is acquired through some branch"
+        # The cost model's own inputs, so a mis-counted feature is visible as itself.
+        assert {"eval_domain", "scan_units", "matches", "n_cards", "n_printings", "residual_tier_ns100"} <= acquire.keys()
+        assert acquire["limit"] == 50
+        assert acquire["offset"] == 0
+        assert acquire["n_cards"] > 0
+        assert acquire["eval_domain"] <= acquire["n_cards"], "candidates cannot exceed the card universe"
 
     def test_explain_analyze_matches_explain_and_times_every_plan(self, engine: QueryEngine) -> None:
         filters = parse_scryfall_query("t:creature")
@@ -1283,12 +1301,48 @@ class TestExplain:
 
         # Same plans, same order, same predicted_ns -- explain_analyze must never
         # diverge from explain (they share the engine's acquire_plan_features step).
-        assert [row["plan"] for row in analyzed] == [row["plan"] for row in explained]
-        assert [row["predicted_ns"] for row in analyzed] == [row["predicted_ns"] for row in explained]
-        for row in analyzed:
+        assert [row["plan"] for row in analyzed["plans"]] == [row["plan"] for row in explained["plans"]]
+        assert [row["predicted_ns"] for row in analyzed["plans"]] == [row["predicted_ns"] for row in explained["plans"]]
+        # The features must agree too, not just the predictions: they are the inputs those
+        # predictions are computed from, and a harness that reads one and times the other would be
+        # silently comparing two different queries.
+        assert analyzed["acquire"]["count_source"] == explained["acquire"]["count_source"]
+        for key in ("eval_domain", "scan_units", "matches"):
+            assert analyzed["acquire"][key] == explained["acquire"][key]
+        for row in analyzed["plans"]:
             # A structurally-applicable plan's fastpath can still decline at runtime
             # (empty trials_ns); otherwise warmups are discarded and exactly
             # num_trials recorded.
             assert row["trials_ns"] == [] or len(row["trials_ns"]) == 3
             assert all(t >= 0 for t in row["trials_ns"])
-        assert any(row["plan"] == "GatheredScan" and len(row["trials_ns"]) == 3 for row in analyzed)
+        assert any(row["plan"] == "GatheredScan" and len(row["trials_ns"]) == 3 for row in analyzed["plans"])
+
+    def test_explain_analyze_counters_track_the_features_that_predict_them(self, engine: QueryEngine) -> None:
+        """Each counter exists to check the feature that is supposed to predict it.
+
+        A mis-counted feature then shows up as itself rather than as a rate that will not calibrate.
+        `t:creature`/card narrows, so the plan that runs must visit candidates and push results, and
+        the phase split must account for the round it reports.
+        """
+        filters = parse_scryfall_query("t:creature")
+        result = engine.explain_analyze(
+            filters=filters,
+            unique="card",
+            prefer="GatheredScan",
+            orderby="edhrec",
+            direction="asc",
+            limit=50,
+            offset=0,
+            num_warmups=1,
+            num_trials=2,
+        )
+        acquire = result["acquire"]
+        gathered = next(row for row in result["plans"] if row["plan"] == "GatheredScan")
+        assert gathered["trials_ns"], "GatheredScan always runs"
+
+        assert gathered["cards_visited"] > 0
+        assert gathered["matches_pushed"] > 0
+        # Card mode returns one row per matching card, so the pushes ARE the distinct-card total.
+        assert gathered["matches_pushed"] <= acquire["n_cards"]
+        # The phase split must account for the round, or a phase's cost is attributed to nothing.
+        assert gathered["ns_round_total"] >= gathered["ns_loop"]
--- a/api/tests/test_engine_unit.py
+++ b/api/tests/test_engine_unit.py
@@ -1348,4 +1348,20 @@ class TestExplain:
         # Card mode returns one row per matching card, so the pushes ARE the distinct-card total.
         assert gathered["matches_pushed"] <= acquire["n_cards"]
         # The phase split must account for the round, or a phase's cost is attributed to nothing.
+        # The three phases are disjoint sub-intervals of the round, so they can only sum to less
+        # than it -- the shortfall is the unmodelled remainder this instrumentation exists to size.
+        # NOT asserted individually as > 0: they are timings, and on a corpus this small
+        # `prepare_candidates` really does complete inside the clock's granularity and report 0.
         assert gathered["ns_round_total"] >= gathered["ns_loop"]
+        phases = gathered["ns_prepare"] + gathered["ns_loop"] + gathered["ns_finish"]
+        assert phases <= gathered["ns_round_total"], "phase split exceeds the round it came from"
+
+        # `result_total` is ground truth for the analyzed run itself, which is the whole reason it
+        # exists: before it, a harness wanting the true cardinality made a SECOND `query()` call and
+        # ASSUMED the two agreed. Check the assumption once here, since every plan is supposed to
+        # return the same total for one query -- that is a correctness invariant, not a measurement.
+        expected_total, _ = _run(engine, "t:creature", unique="card", orderby="edhrec", limit=50)
+        for row in result["plans"]:
+            if not row["trials_ns"]:
+                continue  # declined at runtime, so it produced no total
+            assert row["result_total"] == expected_total, f"{row['plan']} disagrees with query()'s total"
--- a/api/tests/test_engine_unit.py
+++ b/api/tests/test_engine_unit.py
@@ -1328,7 +1328,10 @@ class TestExplain:
         result = engine.explain_analyze(
             filters=filters,
             unique="card",
-            prefer="GatheredScan",
+            # `prefer` is the PRINTING preference, not a plan selector -- `prefer_from_str` maps
+            # anything unrecognised to `Prefer::Default`. explain_analyze times every applicable plan
+            # regardless, so the plan under test is looked up by name from the results below.
+            prefer="default",
             orderby="edhrec",
             direction="asc",
             limit=50,
--- a/card_engine/src/bench_candidate_materialize.rs
+++ b/card_engine/src/bench_candidate_materialize.rs
@@ -1,0 +1,449 @@
+//! Three ways to turn selected postings rows into one sorted candidate list, as a
+//! function of how many candidates come back and how large the candidate space is.
+//!
+//! Context: `arith_tuple_narrow` (and any other arm that unions several postings
+//! rows) has sorted rows but no globally sorted output, because combination order
+//! is not card order. It currently concatenates the selected rows and sorts. Two
+//! alternatives cost nothing extra in information: a k-way merge, since every run
+//! is already sorted, and a bitmap scatter followed by reading the set bits back,
+//! which is sorted by construction.
+//!
+//! The three have different cost shapes, which is why this sweeps two axes:
+//!
+//! - `concat+sort` — O(c log c) in the candidate count, no dependence on the domain.
+//! - `merge` — O(c log k) in the count and the number of runs, no dependence on the
+//!   domain.
+//! - `bitmap+extract` — O(domain/64 + c): a fixed cost proportional to the *space*
+//!   plus a linear pass, so it wins broad results and loses as the domain grows
+//!   relative to the answer.
+//!
+//! Runs are disjoint (each card belongs to exactly one combination) and their ids
+//! are scattered through the domain rather than contiguous, which is the real
+//! situation: a combination's cards sit wherever they sit in store order.
+//!
+//! All three are asserted to produce the identical `Vec<u32>` before anything is
+//! timed.
+//!
+//!     cargo test --release bench_candidate_materialize -- --ignored --nocapture
+//!
+//! Needs no store — the inputs are synthetic so the domain can be varied.
+
+use std::cmp::Reverse;
+use std::collections::BinaryHeap;
+use std::hint::black_box;
+use std::time::Instant;
+
+const ITERS: usize = 60;
+
+/// Real corpus size, so one row of every table is the case the engine actually has.
+const REAL_DOMAIN: usize = 31_508;
+/// Distinct arith-tuple combinations over that corpus; the natural run count there.
+const REAL_RUNS: usize = 564;
+/// The engine's current sorted-vec/bitmap switch, for reference in the output.
+const BITS_PROMOTE_REF: usize = 4_096;
+/// Printing-space domain at the same corpus, the other space a narrowing arm can land in.
+const PRINTING_DOMAIN: usize = 97_206;
+
+fn time_ns(mut kernel: impl FnMut() -> usize) -> f64 {
+    let mut best = u128::MAX;
+    let mut out = 0;
+    for _ in 0..ITERS {
+        let t0 = Instant::now();
+        out = black_box(kernel());
+        best = best.min(t0.elapsed().as_nanos());
+    }
+    black_box(out);
+    best as f64
+}
+
+/// `runs` disjoint sorted id lists drawn from `0..domain`, holding `count` ids in
+/// total, each run's ids scattered rather than contiguous.
+fn make_runs(domain: usize, count: usize, runs: usize) -> Vec<Vec<u32>> {
+    assert!(count <= domain && runs >= 1);
+    // Deterministic LCG: reproducible without a dev-dependency.
+    let mut state = 0x2545_F491_4F6C_DD1Du64;
+    let mut next = move || {
+        state ^= state << 13;
+        state ^= state >> 7;
+        state ^= state << 17;
+        state
+    };
+    // Pick `count` distinct ids by striding with a jitter, then sorting — cheaper
+    // than rejection sampling and still scattered.
+    let stride = domain / count.max(1);
+    let mut ids: Vec<u32> = (0..count)
+        .map(|i| {
+            let base = i * stride;
+            let j = if stride > 1 { (next() as usize) % stride } else { 0 };
+            (base + j).min(domain - 1) as u32
+        })
+        .collect();
+    ids.sort_unstable();
+    ids.dedup();
+    let mut out: Vec<Vec<u32>> = vec![Vec::new(); runs];
+    for id in ids {
+        out[(next() as usize) % runs].push(id);
+    }
+    for r in &mut out {
+        r.sort_unstable();
+    }
+    out
+}
+
+fn concat_sort(runs: &[Vec<u32>], total: usize) -> Vec<u32> {
+    let mut out: Vec<u32> = Vec::with_capacity(total);
+    for r in runs {
+        out.extend_from_slice(r);
+    }
+    out.sort_unstable();
+    out
+}
+
+/// k-way merge over the already-sorted runs. Heap entries are (value, run, pos).
+fn kway_merge(runs: &[Vec<u32>], total: usize) -> Vec<u32> {
+    let mut out: Vec<u32> = Vec::with_capacity(total);
+    let mut heap: BinaryHeap<Reverse<(u32, usize, usize)>> = BinaryHeap::with_capacity(runs.len());
+    for (i, r) in runs.iter().enumerate() {
+        if let Some(&v) = r.first() {
+            heap.push(Reverse((v, i, 0)));
+        }
+    }
+    while let Some(Reverse((v, i, p))) = heap.pop() {
+        out.push(v);
+        if let Some(&nv) = runs[i].get(p + 1) {
+            heap.push(Reverse((nv, i, p + 1)));
+        }
+    }
+    out
+}
+
+fn bitmap_extract(runs: &[Vec<u32>], total: usize, domain: usize) -> Vec<u32> {
+    let mut bits = vec![0u64; domain.div_ceil(64)];
+    for r in runs {
+        for &id in r {
+            bits[id as usize / 64] |= 1u64 << (id % 64);
+        }
+    }
+    let mut out: Vec<u32> = Vec::with_capacity(total);
+    for (w, &word) in bits.iter().enumerate() {
+        let mut word = word;
+        while word != 0 {
+            let b = word.trailing_zeros();
+            out.push((w * 64) as u32 + b);
+            word &= word - 1;
+        }
+    }
+    out
+}
+
+/// The scatter half alone — what the `CardBits` branch above `BITS_PROMOTE` already
+/// pays. Subtracting this from `bitmap_extract` prices the extract pass, which is the
+/// only thing lowering `BITS_PROMOTE` would save.
+fn bitmap_scatter(runs: &[Vec<u32>], domain: usize) -> Vec<u64> {
+    let mut bits = vec![0u64; domain.div_ceil(64)];
+    for r in runs {
+        for &id in r {
+            bits[id as usize / 64] |= 1u64 << (id % 64);
+        }
+    }
+    bits
+}
+
+// ─── Cost model ───────────────────────────────────────────────────────────────
+// Both plans' inputs are known before either structure is allocated: the caller's
+// first pass already sums the selected postings' lengths, and the rows are disjoint
+// (each card belongs to exactly one combination), so that sum is the exact candidate
+// count rather than an upper bound; the domain is `n_cards`. So the choice can be
+// modelled from two numbers already in hand, and only the winning structure ever
+// gets built. These constants are fit against axis A and B above, not assumed.
+
+/// `Vec::with_capacity` plus the run walk, before any comparison work.
+const SORT_FIXED_NS: f64 = 143.0;
+/// pdqsort on `u32`, per candidate — **linear**, not `c·log2 c`.
+///
+/// `sort_unstable` is a full pdqsort, so it is asymptotically `n log n`; this is not a
+/// claim otherwise. But over the three decades that matter here the measured per-element
+/// cost is flat — 5.09, 4.93, 4.92 and 5.06 ns at 1,024 / 4,096 / 16,384 / 31,508 (axis
+/// H) — where an `n log n` term fit to the same data predicts it rising from 3.6 to 5.4.
+/// So the log factor is not observable against per-element memory cost across this range,
+/// and a linear fit is both simpler and more accurate. Fit on the 64 and 1,024 rows of
+/// axis A, which brackets the crossover; it holds to ~6% out to 31,508.
+///
+/// This does NOT license extrapolating past the corpus sizes swept here. The log factor
+/// is real and will eventually show up; re-fit rather than trusting this beyond ~3M.
+const SORT_PER_CAND_NS: f64 = 4.95;
+/// The zeroed `vec![0u64; words]` allocation.
+const BITMAP_FIXED_NS: f64 = 200.0;
+/// Word scan on the extract pass. Fit beyond L1 (0.46–0.53 ns/word across axis B's
+/// 100k–3M domains); inside L1 the real rate is ~0.23, so this over-prices small
+/// domains, which biases the choice toward sort exactly where sort is already fine.
+const BITMAP_PER_WORD_NS: f64 = 0.50;
+/// One scatter store plus one `trailing_zeros` extraction per candidate.
+const BITMAP_PER_CAND_NS: f64 = 0.83;
+
+fn model_sort_ns(count: usize) -> f64 {
+    SORT_FIXED_NS + SORT_PER_CAND_NS * count as f64
+}
+
+/// Both plans are linear in the candidate count, so the crossover has a closed form:
+/// solving `model_sort_ns(c) == model_bitmap_ns(c, domain)` for `c` gives an *affine*
+/// function of the word count. Reported next to the measured crossover so the shape of
+/// the boundary is checked, not just its value at a few domains.
+fn model_crossover(domain: usize) -> f64 {
+    let words = domain.div_ceil(64) as f64;
+    (BITMAP_FIXED_NS - SORT_FIXED_NS + BITMAP_PER_WORD_NS * words) / (SORT_PER_CAND_NS - BITMAP_PER_CAND_NS)
+}
+
+fn model_bitmap_ns(count: usize, domain: usize) -> f64 {
+    BITMAP_FIXED_NS + BITMAP_PER_WORD_NS * domain.div_ceil(64) as f64 + BITMAP_PER_CAND_NS * count as f64
+}
+
+/// Candidate counts bracketing the crossover finely enough to locate it.
+const CROSSOVER_COUNTS: [usize; 12] = [16, 24, 32, 48, 64, 96, 128, 192, 256, 384, 512, 1_024];
+
+/// Fine sweep step for locating the crossover, as a percentage. The coarse
+/// `CROSSOVER_COUNTS` grid steps by ~1.5x, which cannot resolve whether the boundary
+/// collapses to a fixed domain:count ratio — that question needs steps finer than the
+/// ratio drift being tested for.
+const FINE_STEP_PCT: usize = 12;
+/// Consecutive bitmap wins required to call a crossover, so one noisy sample at these
+/// sub-microsecond sizes cannot declare it early.
+const FINE_CONFIRM_STEPS: usize = 2;
+/// Where the fine sweep starts; below this the bitmap has never won at any domain.
+const FINE_START: usize = 32;
+
+/// The measured crossover, located by a fine geometric sweep rather than read off the
+/// coarse grid. Returns the first count with `FINE_CONFIRM_STEPS` consecutive bitmap
+/// wins, or None if the bitmap never wins over the swept range.
+fn fine_crossover(domain: usize, runs_n: usize) -> Option<usize> {
+    let mut count = FINE_START;
+    let mut streak: Vec<usize> = Vec::new();
+    while count <= domain {
+        let runs = make_runs(domain, count, runs_n);
+        let total: usize = runs.iter().map(Vec::len).sum();
+        let sort_ns = time_ns(|| concat_sort(&runs, total).len());
+        let bitmap_ns = time_ns(|| bitmap_extract(&runs, total, domain).len());
+        if bitmap_ns < sort_ns {
+            streak.push(total);
+            if streak.len() >= FINE_CONFIRM_STEPS {
+                return streak.first().copied();
+            }
+        } else {
+            streak.clear();
+        }
+        count = (count * (100 + FINE_STEP_PCT) / 100).max(count + 1);
+    }
+    None
+}
+
+/// Measured vs modelled for one (domain, count) cell, plus the regret from following
+/// the model: how much worse the model's pick is than the better plan, measured.
+struct Cell {
+    count: usize,
+    sort_ns: f64,
+    bitmap_ns: f64,
+    model_picks_bitmap: bool,
+    regret_ns: f64,
+}
+
+fn cells(domain: usize, runs_n: usize) -> Vec<Cell> {
+    CROSSOVER_COUNTS
+        .iter()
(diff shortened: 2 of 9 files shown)


card_engine/src/lib.rs
```diff
@@ -6785,31 +7065,78 @@ fn explain_analyze(
...
+        let acquired = acquire_plan_features(ctx, params, &mut acquire_filter, plane);
+        let acq_dt = t_acq.elapsed().as_nanos() as u64;
+        drop(acquired);
+        // The real routed path, including the dispatch-time re-choose that `run_query_with_plan`
+        // cannot exercise. See AcquireFacts::routed_ns for why this row is the load-bearing one.
+        let mut routed_filter = filter.clone();
+        let t_routed = std::time::Instant::now();
+        let routed = run_query_routed(ctx, params, &mut routed_filter, plane);
+        let routed_dt = t_routed.elapsed().as_nanos() as u64;
+        drop(routed);
```
jbylund: **The routed run contaminates the first plan's `PhaseStats`.**

`run_query_routed` dispatches into `exec_from_candidates` → `exec_gathered_scan` / `run_query_streamed`, both of which publish into `PHASE_STATS`. Nothing clears the cell between here and `take_phase_stats()` in the plan loop below.

So for `idx = round % n`, if that plan is one of the four *uninstrumented* ones (`PrintingRangeScan`, `PrintingCompose`, `PlanePopcountOrder`, `CardRangePopcount`), `take_phase_stats()` returns the **routed run's** counters and timings, and `phases[idx] = stats` stores them. `PhaseStats`' own doc says "Zeros otherwise"; that contract doesn't hold.

Concretely — a plane-acquired `t:creature`/card where `PlanePopcountOrder` is index 0: at round 0 it is credited with GatheredScan's `cards_visited`, `matches_pushed`, `ns_loop`, `ns_finish`, **and** `ns_prepare`. That last one flows straight into `bench_cost_model_agreement.py`'s `prep_frac[p["plan"]]` — a prepare share attributed to a plan that never called `prepare_candidates`.

Second manifestation: because both publishers use `..c.get()` to preserve `paging_taken`, a compose-capable query whose routed run took e.g. `Perm` leaves that label in the cell, and the first plan to publish — say `GatheredScan` — inherits it. `paging_taken`'s doc promises `""` for every other plan.

One line:

```rust
        drop(routed);
        take_phase_stats(); // the routed run publishes too; don't credit it to the first plan
```

card_engine/src/lib.rs
```diff
@@ -6983,7 +7310,13 @@ fn run_query_streamed<'a>(
     counts.resize(cards.len(), 0);
     let have_group_counts = artwork_groups.len() == cards.len();
     let mut total: usize = 0;
+    // Same counters GatheredScan publishes, so the two plans' arms can be checked the same way.
+    // Locals in the loop; published once at the end. See PhaseStats.
+    let (mut n_cards_visited, mut n_printings_scanned, mut n_matches_pushed) = (0u64, 0u64, 0u64);
+    let t_match = std::time::Instant::now();
     for cid in card_ids {
+        n_cards_visited += 1;
+        n_printings_scanned += (u32::from(offsets[cid as usize + 1]) - u32::from(offsets[cid as usize])) as u64;
```
jbylund: **`printings_scanned` is counted differently here than in `exec_gathered_scan`.**

This increments at the *top* of the loop, before the `Tri::False | Tri::Null => continue` below — so it counts printings of every visited card. `exec_gathered_scan` increments the same counter *after* its `continue` (line 6468), so it counts only cards that passed the card-level filter.

The comment two lines up says "Same counters GatheredScan publishes, so the two plans' arms can be checked the same way" — they aren't the same counter. Any query with an inexact narrowing (residual rejects candidates) reports two different `printings_scanned` for identical work, and `scan_units` has one definition, so at most one of the two is the valid comparison.

This is exactly the class of error the PR exists to make visible, so worth getting right before the harness starts fitting on it. Moving the increment below the `continue` matches GatheredScan.

card_engine/src/lib.rs
```diff
@@ -2648,6 +2648,24 @@ impl Candidates {
...
+    /// Which representation the narrowing produced, for `explain` to report. A vec-shaped
+    /// result means some site built a sorted vec — a `collect` + `sort_unstable` ran. A
+    /// bits-shaped one means the narrowing stayed word-wise and never sorted.
+    ///
+    /// Read at the *top* of the narrowing, so it is a proxy rather than a census of sort
+    /// events: `or_all` can scatter a vec-shaped arm into bits, hiding a sort that did
+    /// happen, and a bits-shaped arm can be extracted into a vec by `and_all`. Exact
+    /// per-site accounting needs the shared materialization helper that
+    /// docs/issues/local-engine-candidate-materialize.md proposes.
+    fn repr_label(&self) -> &'static str {
```
jbylund: This was inserted *between* `into_bits`' doc comment and `into_bits`. The result: "The set as a bitmap over an n-element domain…" (2649–2650) now documents `repr_label`, and `into_bits` at 2669 is undocumented.

Move `repr_label` above line 2649, or below `into_bits`.

card_engine/src/lib.rs
```diff
@@ -6290,6 +6332,93 @@ fn exec_streamed_select<'a>(
 /// P4 executor: the universal gathered per-card loop + `select_page`. Runs any
 /// query (printing-keyed orderbys, stores without permutations, or anything the
 /// other plans decline). Caller has run `prepare_candidates`.
+/// Per-query execution counters and coarse phase timings, for checking the cost model against what
```
jbylund, line 6389: Same doc-detachment as `repr_label`: the "P4 executor: the universal gathered per-card loop + `select_page`…" comment at 6332–6334 now heads `PhaseStats`, and `exec_gathered_scan` at 6422 is undocumented — with ~90 lines of thread-local machinery between the doc and the function it describes.

Both are mechanical, but in a file this heavily commented the docs are the interface.

card_engine/src/cost.rs
```diff
@@ -179,6 +180,58 @@ pub(crate) const RANGE_SCATTER_PER_PRINTING_NS: f64 = 0.36;
...
+//
+// So it is REPORTED BY `explain`, NOT ADDED TO `plan_cost`, for three reasons: it is validated as
+// too small to help; it is identical for the plans that actually compete (`StreamedSelect` and
+// `GatheredScan` call the same `prepare_candidates`), so it cannot change an argmin; and its real
+// purpose is to price the bitmap-versus-sort question in
+// docs/issues/local-engine-candidate-materialize.md, which is what it does measure exactly.
+
+/// `Vec::with_capacity` plus the run walk, before any comparison work
+/// (`bench_candidate_materialize`, axis A).
+const MATERIALIZE_SORT_FIXED_NS: f64 = 143.0;
```
jbylund: **Duplicated constants with nothing tying them together.**

`143.0` / `4.95` appear here as `MATERIALIZE_SORT_FIXED_NS` / `MATERIALIZE_SORT_PER_CAND_NS`, and again as `SORT_FIXED_NS` / `SORT_PER_CAND_NS` at `bench_candidate_materialize.rs:161`/`:174`. This is the same drift shape `PhaseStats::paging_taken`'s own doc calls out — "the Python cost mirror which silently drifted from `cost.rs` for two revisions."

The bench is a `cfg(test)` module inside the crate, so making these `pub(crate)` and having the bench `use` them costs nothing.

Their justifying docs already disagree: this one says "4.39 ns at 1,024 rising only to 5.09 at 31,508"; the bench says "5.09, 4.93, 4.92 and 5.06 ns at 1,024 / 4,096 / 16,384 / 31,508". One of the two is stale.

api/tests/test_engine_unit.py
```diff
@@ -1283,12 +1301,48 @@ def test_explain_analyze_matches_explain_and_times_every_plan(self, engine: Quer
...
+
+        A mis-counted feature then shows up as itself rather than as a rate that will not calibrate.
+        `t:creature`/card narrows, so the plan that runs must visit candidates and push results, and
+        the phase split must account for the round it reports.
+        """
+        filters = parse_scryfall_query("t:creature")
+        result = engine.explain_analyze(
+            filters=filters,
+            unique="card",
+            prefer="GatheredScan",
```
jbylund: `prefer` is the *printing* preference, not a plan selector — `prefer_from_str` maps anything unrecognized to `Prefer::Default`, so this is silently just `"default"`.

The test passes, and looking GatheredScan up by name in the results is the right approach, but the argument implies a plan-forcing mechanism that doesn't exist on this entry point and will mislead the next reader. Drop it or use `"default"`.

scripts/bench_cost_model_agreement.py
```diff
@@ -0,0 +1,184 @@
...
+            "orderby": rng.choice(ORDERBYS),
+            "direction": rng.choice(("asc", "desc")),
+            "limit": rng.choice(LIMITS),
+            "offset": rng.choice(OFFSETS),
+        }
+        try:
+            kw["filters"] = parse_scryfall_query(q)
+            acq = engine.explain(**kw)["acquire"]
+            res = engine.explain_analyze(prefer="default", num_warmups=NUM_WARMUPS, num_trials=NUM_TRIALS, **kw)
+        except Exception:  # noqa: BLE001 - a rejected query is a skipped sample
```
jbylund: **Swallowing every failure with no reason breakdown undercuts the module docstring's own argument.**

The docstring's whole case is that sampling bias hides the cells most likely to be wrong. But if `explain_analyze` starts raising for, say, all `unique=artwork` queries, this prints a healthy-looking table over two thirds of the intended space, with only an aggregate skip count to hint at it.

A `collections.Counter` of `type(exc).__name__`, printed alongside the skip total in `report`, is three lines and closes the gap.

card_engine/src/cost.rs
```diff
@@ -179,6 +180,58 @@ pub(crate) const RANGE_SCATTER_PER_PRINTING_NS: f64 = 0.36;
...
+/// alternative that doc proposes; if that ships, this needs the domain term as well.
+///
+/// The match below is deliberately NOT `PhysicalPlan::materializing()`, which means something
+/// else — "runnable off a materialized prep", and so includes `PlanePopcountOrder`, which reads
+/// the plane bitmap directly and builds no candidate list. Charging it here would invert exactly
+/// the plane-against-materializing comparison this term exists to inform.
+pub(crate) fn materialize_cost(plan: PhysicalPlan, f: &PlanFeatures) -> f64 {
+    match plan {
+        PhysicalPlan::StreamedSelect | PhysicalPlan::GatheredScan => {
+            MATERIALIZE_SORT_FIXED_NS + MATERIALIZE_SORT_PER_CAND_NS * f64::from(f.eval_domain)
```
jbylund, line 238: Nit: `eval_domain` is only the candidate count under a `Candidates` acquire. Under `Prep::Range` the two materializing plans are estimated unnarrowed (`eval_domain = n_cards`), so `materialize_ns` is a number with no referent there.

The section comment above is clear that this isn't routing-load-bearing, but the field is emitted to Python unqualified — worth a sentence on `PlanEstimate::materialize_ns` so a harness doesn't group range-acquired rows in with candidate-acquired ones.

card_engine/src/lib.rs
```diff
@@ -6714,6 +6870,98 @@ fn run_query_with_plan<'a>(
...
+    ///
+    /// Needed because a candidate *count* in a given band does not imply a sort was paid to
+    /// reach it: a plane AND'd with a range lands at the same count without sorting anything.
+    ///
+    /// A top-of-narrowing proxy, not a per-site census — see `Candidates::repr_label`.
+    pub(crate) narrowed_repr: &'static str,
+    /// Candidate cards the walk would iterate (`PlanFeatures::eval_domain`). For a
+    /// `"candidates"` query this IS the materialized candidate count; compare it
+    /// against `n_cards` — equal means nothing narrowed.
+    pub(crate) eval_domain: u32,
```
jbylund: Nit: `eval_domain`, `n_cards` and `matches` are carried both here and inside `feats`, which is why the `feats.clone()` at construction is needed. `acquire_facts_to_pydict` reads these three from the top level and everything else from `feats` — collapsing to just `feats` would drop both the clone and the duplication, with no change to the emitted dict.

card_engine/src/bench_candidate_materialize.rs
```diff
@@ -0,0 +1,449 @@
...
+    }
+    black_box(out);
+    best as f64
+}
+
+/// `runs` disjoint sorted id lists drawn from `0..domain`, holding `count` ids in
+/// total, each run's ids scattered rather than contiguous.
+fn make_runs(domain: usize, count: usize, runs: usize) -> Vec<Vec<u32>> {
+    assert!(count <= domain && runs >= 1);
+    // Deterministic LCG: reproducible without a dev-dependency.
```
jbylund: Nit: this is an xorshift64, not an LCG.

jbylund (commented):
Reviewed the instrumentation, the bench, the issue doc and the harness.

The framing is right — landing measurement before the changes it will judge — and the evidence in the description is honest (reporting both benchmark orderings rather than the flattering one is the correct call). No in-tree consumers other than the tests and the new script depend on the old list shape, so the `PyList` → `PyDict` break is contained.

Two defects worth fixing before merge, both in the measurement this PR exists to enable:

1. The routed run at the top of each `explain_analyze` round publishes into `PHASE_STATS`, and nothing clears it before the plan loop — so the first plan in the rotation inherits the routed run's counters whenever that plan is one of the four uninstrumented ones. `bench_cost_model_agreement.py` consumes exactly those fields.
2. `printings_scanned` is incremented on opposite sides of the `card_pass` `continue` in the two executors, so the two plans report different counts for the same query. `scan_units` has one definition, so at most one of them is the valid comparison.

The rest are inline and cheap.

### Things that are right

- `Prep::Range(&'static str)` is the correct fix for the reported problem, and the fuzz test asserting the label is one of five *known* strings (rather than just non-empty) will catch a new branch that forgets to name itself.
- `note_paging_taken` covering all four fastpath exits including the two early returns — the decline paths are the ones that would otherwise silently vanish.
- `facts.acquire_ns.clear()` before re-sampling in-loop, with the rationale written down.
- Asserting all three materialization strategies produce byte-identical `Vec<u32>` before timing anything.
- The issue doc's **"Do not build the merge"** — a negative result stated as the useful half of the finding, and the explicit scoping of what the production-side bench cannot settle about `BITS_PROMOTE`.

One process note, not a defect: `explain_analyze` now does an extra `acquire_plan_features` plus a full `run_query_routed` per round on top of the per-plan loop, so a given `--seconds` budget buys meaningfully fewer samples than it did before. Worth a line in the harness docstring so a later comparison against an older run isn't misread as a sampling regression.

card_engine/src/cost.rs
jbylund: Half of this is fixed, half is now explicitly deferred — leaving the thread open for the second half.

**Fixed:** the bench `use`s the constants from `cost.rs` (`bench_candidate_materialize.rs:163`), so the values can no longer drift.

**Still open:** the stale figure. The new doc at `cost.rs:211-215` acknowledges the disagreement rather than adjudicating it — "One of the two is stale; adjudicating it needs a re-run on a machine that is not clock-throttled, so it is flagged rather than silently reconciled."

Flagging over silently reconciling is the right call, and sharing the value means only prose can drift now. But `MATERIALIZE_SORT_PER_CAND_NS = 4.95` is the sole basis for the issue doc's "keep concat+sort below ~64 candidates" recommendation, and it is now emitted to Python as `materialize_ns`. The bench is committed and takes seconds to run, so one pass on a quiesced machine settles it — worth doing before a follow-up PR leans on the constant to justify a behavior change.

card_engine/src/lib.rs
```diff
@@ -6287,6 +6329,93 @@ fn exec_streamed_select<'a>(
...
+    /// the cost model predicted it would.
+    ///
+    /// `compose_paging_with_total` reimplements a decision the fastpath makes independently -- it
+    /// recomputes the permutation lookup, the orderby-walk test and the decline conditions on its
+    /// own -- and nothing checked that the two agree. That is the same shape as the Python cost
+    /// mirror which silently drifted from `cost.rs` for two revisions. Reporting the real branch
+    /// turns the assumption into something a harness can assert.
+    ///
+    /// `""` for every other plan, and for a compose run that never reached the paging step.
+    pub(crate) paging_taken: &'static str,
```
jbylund: **`paging_taken` and `result_total` have no assertion and no consumer — anywhere.**

This doc makes the strongest argument in the PR: `compose_paging_with_total` independently reimplements the fastpath's branch decision, "nothing checked that the two agree… That is the same shape as the Python cost mirror which silently drifted from `cost.rs` for two revisions. Reporting the real branch turns the assumption into something a harness can assert."

Nothing asserts it. `paging_taken` and `result_total` appear in zero tests — not `tests.rs`, not `test_engine_unit.py` — and `bench_cost_model_agreement.py` never reads either field. The drift this exists to catch is still undetected; the PR ships the sensor without wiring it up.

Two cheap closures:
- In `bench_cost_model_agreement.py`, compare `acq["compose_paging"]` against the compose row's `paging_taken` and count disagreements as their own cell. That is the harness assertion this doc promises.
- One `assert result_total == <query()'s total>` in `test_engine_unit.py`, which is the whole reason the field was added ("a harness wanting the true cardinality made a SECOND `query()` call… so agreement was assumed rather than observed").

scripts/bench_cost_model_agreement.py
```diff
@@ -0,0 +1,193 @@
...
+            continue
+        agr.sampled += 1
+        for p in res["plans"]:
+            if not p["trials_ns"] or p["predicted_ns"] <= 0:
+                continue
+            measured = min(p["trials_ns"])
+            agr.ratios[p["plan"], acq["count_source"]].append(measured / p["predicted_ns"])
+            agr.by_unique[p["plan"], unique].append(measured / p["predicted_ns"])
+            if p["ns_round_total"]:
+                agr.prep_frac[p["plan"]].append(p["ns_prepare"] / p["ns_round_total"])
```
jbylund: **The module docstring promises a split this doesn't compute.**

The docstring says the script reports "what fraction of the time is `prepare_candidates` — which no cost term currently describes and which is 21-33% of a range-acquired query against 7-10% of a plane-acquired one."

`prep_frac` is keyed by plan alone, and `report()` prints one row per plan — there is no acquire-branch dimension in the table at all. Compounding it: `ns_prepare` is only ever nonzero for the two plans that go through `timed_prepare_candidates`, so this table can never have more than two rows and can never show the range-vs-plane contrast the docstring quotes as its motivating number.

`acq["count_source"]` is already in scope on the line above:

```python
agr.prep_frac[p["plan"], acq["count_source"]].append(p["ns_prepare"] / p["ns_round_total"])
```

That also lets `report()` reuse `summarise`'s two-key layout instead of its own loop.

card_engine/src/lib.rs
```diff
@@ -6714,6 +6870,97 @@ fn run_query_with_plan<'a>(
 pub(crate) struct PlanEstimate {
     pub(crate) plan: PhysicalPlan,
     pub(crate) predicted_ns: f64,
+    /// `cost::materialize_cost` for this plan — the candidate-production term `plan_cost` omits,
+    /// reported but deliberately NOT added to `predicted_ns`. `0.0` for plans that build no
+    /// candidate list. See `cost.rs`'s "Candidate materialization" section for why it stays out
+    /// of the routing decision.
+    /// Modelled `collect` + `sort_unstable` cost for this plan's candidate list.
```
jbylund: Two doc paragraphs concatenated here — looks like the fix for the `eval_domain`-has-no-referent thread was appended without deleting what it replaced.

The first block (6873-6876) ends "See `cost.rs`'s 'Candidate materialization' section for why it stays out of the routing decision", then this line restarts with "Modelled `collect` + `sort_unstable` cost for this plan's candidate list." Both describe the same field from the top.

The second block is the better one — it carries the `Prep::Range` caveat that thread asked for. Drop 6873-6876 and keep 6877 onward, adding the `cost.rs` pointer to it.

card_engine/src/lib.rs
```diff
@@ -6746,7 +7012,17 @@ fn explain(ctx: &QueryCtx, params: &QueryParams, filter: &mut FilterExpr, plane:
...
     pub(crate) plan: PhysicalPlan,
     pub(crate) predicted_ns: f64,
+    /// Both carried through from `PlanEstimate` unchanged — see its fields. `picked` in particular
+    /// is the router's choice, which is NOT necessarily the fastest `trials_ns`: that difference is
+    /// the whole point of docs/issues/local-engine-plan-misselection.md.
+    pub(crate) materialize_ns: f64,
+    pub(crate) picked: bool,
     pub(crate) trials_ns: Vec<u64>,
+    /// Execution counters and coarse phase timings from the last recorded run of this plan, when the
+    /// executor is instrumented (`GatheredScan` today). Zeros otherwise. See `PhaseStats`: the
```
jbylund: Stale as of this PR: `StreamedSelect` is instrumented too — `run_query_streamed` publishes from all three of its return points via the `publish` closure. A reader following this doc will wrongly discard StreamedSelect rows as zeros, which is half the instrumented surface.

The plans that genuinely report zeros are `PrintingRangeScan`, `PrintingCompose`, `PlanePopcountOrder` and `CardRangePopcount`. Worth naming them explicitly rather than naming the instrumented set, since that list is the one a consumer has to filter on.

Related, and worth a sentence here: because `explain_analyze` always sets `ns_round_total` but only instrumented executors set `ns_loop`, `ns_round_total > 0 && ns_loop == 0` is the signal for "uninstrumented" as distinct from "did no work". Otherwise a consumer cannot tell the two apart.

card_engine/src/lib.rs
```diff
@@ -6287,6 +6329,93 @@ fn exec_streamed_select<'a>(
...
+    /// own -- and nothing checked that the two agree. That is the same shape as the Python cost
+    /// mirror which silently drifted from `cost.rs` for two revisions. Reporting the real branch
+    /// turns the assumption into something a harness can assert.
+    ///
+    /// `""` for every other plan, and for a compose run that never reached the paging step.
+    pub(crate) paging_taken: &'static str,
+}
+
+thread_local! {
+    static PHASE_STATS: std::cell::Cell<PhaseStats> = const { std::cell::Cell::new(PhaseStats {
```
jbylund, line 6657: `PHASE_STATS` is never cleared on the production path — only `explain_analyze` calls `take_phase_stats()`. So a live worker thread accumulates a `paging_taken` / `result_total` from whatever query it last ran, and every subsequent `..c.get()` publish in `exec_gathered_scan` / `run_query_streamed` re-inherits it.

Harmless today: nothing outside `explain_analyze` reads the cell, and `explain_analyze` clears it before each measurement. But the `..c.get()` idiom exists precisely so one publisher can inherit another's field, which makes "the cell holds stale data from an unrelated query" a trap for the next person who adds a reader.

Either have `run_query_routed` drop the stats on the way out, or say plainly on this declaration that the cell is only meaningful between a `take_phase_stats()` and the next publish, and that anything reading it outside `explain_analyze` is reading an arbitrary earlier query.

card_engine/src/cost.rs
```diff
@@ -60,6 +60,7 @@ use super::*;
 /// Cheap, per-query features the cost model consumes, built once per query by
 /// `run_query_routed`'s `acquire` step. All counts are exact or cheap-exact (plane
 /// popcount / range `k` / candidate count), never estimated.
+#[derive(Clone)]
```
jbylund: This derive is now dead. It was added for the `feats.clone()` that `AcquireFacts` needed when it duplicated `eval_domain`/`n_cards`/`matches` at the top level; collapsing that to just `feats` (which was the right fix) means `explain` now moves the value — `feats, // moved: eval_domain/n_cards/matches live here and nowhere else`. Nothing in the crate clones a `PlanFeatures`.

Clippy won't flag an unused derive, so it will sit here indefinitely. Drop it.

card_engine/src/bench_candidate_materialize.rs
```diff
@@ -0,0 +1,438 @@
...
+
+fn model_sort_ns(count: usize) -> f64 {
+    SORT_FIXED_NS + SORT_PER_CAND_NS * count as f64
+}
+
+/// Both plans are linear in the candidate count, so the crossover has a closed form:
+/// solving `model_sort_ns(c) == model_bitmap_ns(c, domain)` for `c` gives an *affine*
+/// function of the word count. Reported next to the measured crossover so the shape of
+/// the boundary is checked, not just its value at a few domains.
+fn model_crossover(domain: usize) -> f64 {
```
jbylund, line 189: Nit, but sharpened by the constants now being imported rather than local: the divisor `SORT_PER_CAND_NS - BITMAP_PER_CAND_NS` is unguarded. It is 4.12 today, but `SORT_PER_CAND_NS` is exactly the value flagged as possibly stale and due a re-fit — and a re-fit below `0.83` silently yields a *negative* crossover that the table prints as if it meant something. A `debug_assert!` on the sign, or returning `None` when the bitmap is not asymptotically cheaper per candidate, makes the model's precondition explicit.

Two smaller ones in the same file:
- `use super::cost::{…}` at line 163 sits mid-file after several `fn` definitions. The comment above it is the valuable part and belongs where the constants are consumed; the `use` itself reads oddly there.
- `fine_crossover` allocates a `Vec<usize>` purely to count a streak. `Option<usize>` for the first hit plus a counter is clearer and drops the allocation.

jbylund (commented):
Re-reviewed against `adf27a1`. Nine of the ten open threads are genuinely fixed and I've resolved them; the constants thread stays open for its stale-figure half (replied there).

Verified on this HEAD: `cargo check --tests` and `cargo clippy --tests` clean, `ruff check` clean, `pytest -k explain` 4 passed. I also traced the thread-local plumbing for the failure modes it invites and found no live bug — all three `run_query_streamed` return points publish, `PENDING_PREPARE_NS` cannot leak (both consumers unconditionally reach a `replace(0)`), the two executors' `printings_scanned` now agree, and `materialize_cost`'s match is exhaustive so a new plan is a compile error rather than a silent `0.0`. No caller outside the PR is broken by the list→dict change.

What follows is what I still think is open, most-substantive first.

PR #872, opened 2026-08-07
Parsing: Accept Result-Shape Directives and the ci Identity Alias

daveycodez:
## What

*(Slimmed to the parser half per review — the rewrite-vocabulary entries now live in their own PR for the quick-merge path you suggested.)*

Two Scryfall-compatibility gaps hit by a client that forwards Scryfall-shaped query strings verbatim:

1. **Result-shape directives** — Scryfall accepts `sort:edhrec`, `order:name`, `direction:asc`, `prefer:oldest` *inside* the query string; they affect presentation, never membership ([example](https://scryfall.com/search?q=t%3Aplaneswalker+sort%3Aedhrec)). Previously these were parse errors here. Both parsers now consume a directive to `TrueNode`, and a new `prune_true_operands` pass in the rewrite pipeline removes true operands from compounds, so `t:goblin sort:edhrec` parses to a tree **equal** to `t:goblin` (no vestigial `AND TRUE` in SQL; the engine already had `FilterExpr::True`). A dangling `sort:` stays a parse error, and words merely *starting* with a directive ("sorting") keep their name reading.

2. **`ci` alias** — Scryfall treats `ci` as a color-identity alias (`ci<=bg` == `id<=bg`). One-word addition to the identity alias list.

## Validation

- `api/parsing/tests/`: full suite green including new `test_result_directives.py` (both parsers, directive/equivalent SQL equality, parity, the dangling-directive error, the prefix-word case, `ci` ≡ `id` in both parsers)
- `ruff check` and `ruff format --check` clean

### DIFF
--- a/api/parsing/rewrite.py
+++ b/api/parsing/rewrite.py
@@ -25,6 +25,7 @@ from api.parsing.nodes import (
     Query,
     RegexValueNode,
     StringValueNode,
+    TrueNode,
     flatten_nested_operations,
 )
 
@@ -200,10 +201,46 @@ def expand_derived_predicates(query: Query) -> Query:
     return flatten_nested_operations(Query(root))
 
 
+def _prune_true(node: QueryNode) -> QueryNode:
+    """Return `node` with always-true leaves pruned; the original object when nothing changed."""
+    cls = node.__class__
+    if cls is AndNode:
+        ops = [_prune_true(op) for op in node.operands]
+        kept = [op for op in ops if not isinstance(op, TrueNode)]
+        if not kept:
+            return TrueNode()
+        if len(kept) == 1:
+            return kept[0]
+        return AndNode(kept) if kept != list(node.operands) else node
+    if cls is OrNode:
+        ops = [_prune_true(op) for op in node.operands]
+        if any(isinstance(op, TrueNode) for op in ops):
+            return TrueNode()
+        return OrNode(ops) if ops != list(node.operands) else node
+    if cls is NotNode:
+        inner = _prune_true(node.operand)
+        return NotNode(inner) if inner is not node.operand else node
+    return node
+
+
+def prune_true_operands(query: Query) -> Query:
+    """Drop always-true leaves from compounds, so a result-shape directive leaves no residue.
+
+    A directive like `sort:edhrec` parses to the always-true node; without this pass a query
+    carrying one would serialize with a vestigial `AND TRUE`, making `t:goblin sort:edhrec`
+    compare unequal to `t:goblin` despite filtering identically. An Or containing an
+    always-true operand is itself always true.
+    """
+    root = _prune_true(query.root)
+    if root is query.root:
+        return query
+    return Query(root)
+
+
 # The post-parse rewrite pipeline, applied in order at the shared parse seam. Add future AST
 # rewrites to this tuple — both parsers call `rewrite_query`, so a new pass lands in exactly one
 # place and is guaranteed identical treatment across parsers (enforced by test_parser_parity).
-_REWRITE_PASSES = (expand_derived_predicates, lower_literal_regexes)
+_REWRITE_PASSES = (expand_derived_predicates, lower_literal_regexes, prune_true_operands)
 
 
 def rewrite_query(query: Query) -> Query:
--- a/api/parsing/db_info.py
+++ b/api/parsing/db_info.py
@@ -76,7 +76,7 @@ DB_COLUMNS = [
     FieldInfo(
         db_column_name="card_color_identity",
         field_type=FieldType.JSONB_OBJECT,
-        search_aliases=["color_identity", "coloridentity", "id", "identity"],
+        search_aliases=["color_identity", "coloridentity", "id", "identity", "ci"],
         parser_class=ParserClass.COLOR,
     ),
     FieldInfo(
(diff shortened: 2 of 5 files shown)


api/parsing/rewrite.py
```diff
@@ -77,6 +78,21 @@
...
+    # 2026-08-07): lower by the corpus's absent digital/funny legends, and by
+    # eligibility text printed on a back FACE ("... can be your commander")
+    # until face text is searchable (the DFC priority in the README).
+    ("is", "commander"): '(t:legendary t:creature) or o:"can be your commander"',
+    # Land cycles whose oracle wording is an EXACT signature -- each matched
+    # 10/10 against api.scryfall.com on 2026-08-07. The other cycles
+    # (checkland, painland, scryland, ...) share their wording with unrelated
+    # lands, so a text signature over-matches; they are deliberately absent
+    # rather than silently wrong, pending the data-driven card_is_tags path.
+    ("is", "slowland"): 't:land o:"tapped unless you control two or more other lands"',
```
jbylund: Do we want `otag:cycle-slowland` / `otag:cycle-bondland` / `otag:cycle-tangoland`? I think that's partly a what does this mean to scryfall question. I don't use is tags much and as far as I know they're not in a scryfall bulk data dump. The tooling is present to import from scryfall, but it's expensive. One thing I could do is import/re-export is tags to s3 which should catch them all/be cheap.

daveycodez: Adopted — you're right that the cycle tags are the better substrate, and it turned out bigger than the three entries you commented on: with ancestor propagation the parent slugs resolve exactly, so I've reworked all the cycle predicates onto `otag:cycle-*` and added nine more that text signatures couldn't express safely (checkland and painland over-matched badly as text — 42 and 25 vs Scryfall's 10 each). All twelve mappings are count-validated EXACT against api.scryfall.com as of today; the three whose tag taxonomy doesn't line up with Scryfall's `is:` membership (filterland 10 vs 22, gainland 25 vs 15, triome 5 vs 10) stay absent rather than approximate. On the S3 re-export idea: generous offer, but for our deployment we'd rather keep every input instance-buildable (Scryfall bulk in, nothing depending on another instance's bucket or cadence) — the import tooling you mentioned running per-instance seems like the right shape if `is:` data is ever needed beyond these rewrites; happy to discuss on #874.

api/parsing/rewrite.py
```diff
@@ -77,6 +78,34 @@
     # looser form also catches Crawling Barrens ("they become a 0/0 ..."). The "still a land"
     # clause is what keeps false positives at zero.
     ("is", "manland"): "t:land o:become o:creature o:/still a.* land/",
+    ("is", "creatureland"): "t:land o:become o:creature o:/still a.* land/",
+    # Commander eligibility: legendary creatures, plus cards whose rules text
+    # grants it outright. vs api.scryfall.com is:commander (3677 on
+    # 2026-08-07): lower by the corpus's absent digital/funny legends, and by
+    # eligibility text printed on a back FACE ("... can be your commander")
+    # until face text is searchable (the DFC priority in the README).
+    ("is", "commander"): '(t:legendary t:creature) or o:"can be your commander"',
```
jbylund: Can non-creature legendaries be your commander - they come back in scryfall's `is:commander` search?

https://scryfall.com/card/acr/68/adrestia

```
(t:legendary AND (toughness>-1 OR t:background)) OR o:"can be your commander"
```

the toughness>-1 is a hack to catch creatures/vehicles/spacecraft at once, I suppose I'd also be fine with

```
(t:legendary (t:creature or t:vechicle or t:spacecraft or t:background)) OR ...
```

but I'm slightly less confident that there won't be one more type added to that than that the next type added will have toughness>=0, but it won't be that big of a churn either way.

jbylund (commented):
The ticket I was thinking of that might be worth reading was #713.

The part that I need to review in a little more depth is the TrueNode pruning and the parser changes - I think the adding more aliases to the rewrite vocab is pretty straightforward, so if you want to split I can ship/merge the one part more quickly.

api/parsing/rewrite.py
daveycodez: Adopted in #889 (the entries now live there per your split suggestion) with your toughness form — spelled `toughness>=0` since the parser rejects negative literals; `>-1` and `>=0` measure identical on Scryfall (3720 both, `*` compares as 0 either way). Diffing your expansion against `is:commander` turned up something worth knowing for 00713: the +43 over-catch split into back-face legendaries (the face-evaluation cluster your doc already names) and **the commander banlist** — Griselbrand, Golos, Emrakul and Erayo were all in the diff, so Scryfall's `is:commander` excludes banned cards while still including 329 casual not-legal legends (`is:commander -legal:commander`). Shipped as `((t:legendary (toughness>=0 or t:background)) or o:"can be your commander") -banned:commander` with the residual documented at the entry. One caution from trying to validate the exact count server-side: appending `-banned:commander` to the probe on Scryfall *raised* its count by ~376 — negated legality terms interact with their default filters in the undocumented way your is:bear notes describe, so the banlist exclusion is validated by the diff membership rather than count reproduction.

jbylund: > parser rejects negative literals; >-1

sheepish

#891

I could've sworn that worked at some point.

PR #877, opened 2026-08-07
API+Engine: Expose layout, cmc, rarity, color_identity and legalities as Result Fields

daveycodez:
## What

Adds five card-data fields to the `fields=` vocabulary, on **both serving paths** with matching Scryfall JSON shapes:

| field | shape | source |
|---|---|---|
| `layout` | text (`"normal"`, `"transform"`, …) | `card_layout` / interned layout id |
| `cmc` | integer | `cmc` column / archived `Option<u8>` |
| `rarity` | text (`"common"`…`"bonus"`) | `card_rarity_text` / Rust mirror of `magic.rarity_int_to_text` |
| `color_identity` | WUBRG-ordered letter list | JSONB object reshaped post-query (SQL) / bitmap decode (engine) |
| `legalities` | `{format: status}` object | raw JSONB (SQL) / packed-u64 decode through the format registry (engine) |

`DEFAULT_RESULT_FIELDS` is untouched — the usual 9 stay the default; these are opt-in.

## Notes on the two shapes that needed real decisions

- **color_identity**: the stored JSONB is an object (`{"R": true}`); Scryfall's JSON is a list. The SQL path reshapes in `_search_sql` (a tiny `_identity_letters` helper), the engine decodes its `u8` bitmap directly, and both emit Scryfall's canonical WUBRG order so the paths agree by construction — per the "matching semantics" contract on `RESULT_FIELD_COLUMNS`.
- **legalities**: the engine decodes through the format-shift registry (every known format, absent bits → `"not_legal"`, exactly round-tripping `jsonb_obj_to_legality_bits`), and consults the printing-level word only for the ~556 `legality_divergent` cards — the same rule the legality filters already use.

## Why

We're moving a production MTG search service onto Sylvan Librarian as its primary card-search backend. The service runs its own defense-in-depth filtering on every returned card (color identity, format legality, layout playability, rarity/cmc constraints from user-pinned filters), which today forces a secondary per-card fetch for data the corpus already holds. With these five fields, one `/search` call carries everything downstream filtering needs.

## Validation

- `cargo test`: 155 passed; new helpers covered via the field tests
- `api/tests/test_engine_unit.py`: 161 passed, including new tests for the five shapes against the fixture corpus and WUBRG ordering on multicolor identities
- `api/tests/test_api_resource.py`: 109 passed, including `_identity_letters` unit tests
- `ruff check` + `ruff format` clean

### DIFF
--- a/api/api_resource.py
+++ b/api/api_resource.py
@@ -285,7 +285,28 @@ RESULT_FIELD_COLUMNS: dict[str, str] = {
     "scryfall_id": "scryfall_id",
     "price_usd": "price_usd",
     "prefer_score": "prefer_score",
+    # Card-data fields consumers need to run their own downstream filtering
+    # (Scryfall JSON names and shapes): layout and rarity are plain text,
+    # cmc an integer, legalities the imported {format: status} object, and
+    # color_identity a WUBRG-ordered letter list (the raw column is a JSONB
+    # object -- _search_sql reshapes it; the engine emits the list directly).
+    "layout": "card_layout",
+    "cmc": "cmc",
+    "rarity": "card_rarity_text",
+    "color_identity": "card_color_identity",
+    "legalities": "card_legalities",
 }
+
+# Scryfall's canonical color order, used to reshape identity objects into lists.
+_COLOR_ORDER: tuple[str, ...] = ("W", "U", "B", "R", "G", "C")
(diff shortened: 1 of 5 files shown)


jbylund:
Thank you merged as #959

PR #879, opened 2026-08-07
Parsing+Engine: Numeric Color-Identity Comparisons (id>=3)

daveycodez:
## What

Scryfall accepts a bare integer on the color fields, meaning **"number of colors in the field"** — this PR adds that syntax to both parsers, the SQL generator, and the Rust engine, for **both** `id`/`identity` and `c`/`color`:

| Query | Meaning |
|---|---|
| `id=2` | identity has exactly two colors |
| `id>=3` | identity has three or more colors |
| `id<2` | fewer than two (colorless counts as zero) |
| `c=2` | exactly two printed colors |
| `id:2` | same as `id=2` — `:` with a number is equality |

Previously both parsers rejected these (`Invalid color value '2'`).

### Semantics verified against the live Scryfall API (2026-08)

- `id>=3` → 1056 total_cards (`unique=cards`)
- `id=2` and `id:2` → 4775 each (so `:` = `=` for numbers, unlike the subset semantics letters get)
- `c=2` and `c:2` → 3811 each
- `id<2` → 27764

## Design shipped

**Full support in all three layers** (not the SQL-fallback variant):

- **Parsers** (`hand_parser.py`, `pyparsing_based.py`): a bare integer is a valid COLOR-class value and produces a raw `NumericValueNode` instead of a color-set value. Floats (`id=2.5`) and mixed tokens (`id:2rr`) stay rejected, identically in both parsers — parser parity holds on both the success and the error side.
- **SQL** (`card_query_nodes.py`): the count comparison reuses the indexed mask expression the subset queries already use — `magic.color_identity_mask(col) = ANY(masks)` with the mask set restricted to values whose popcount satisfies the comparison — so `id`-count queries ride the existing expression index (`c`-count uses the same shape, unindexed but correct, since the function reads only the five WUBRG keys).
- **Rust engine** (`filter.rs`): new `FilterExpr::ColorCountCmp { field, op, count }` variant — popcount of the WUBRG bits (C bit masked off, mirroring the SQL function) vs the queried count. It's total and two-valued like `ColorCmp`, so it's registered in `is_total_two_valued` (negation stays the exact complement) and classified in every exhaustive match (`verify_cost_tier`, `leaf_compares_printing_field`, `has_printing_varying_leaf`); the estimator returns the sound "unknown" (no plane compilation yet).

**`produces` deliberately refuses numbers** with a clear `ValueError`/engine error in all layers: `produced_mana`'s `C` key is a genuine producible value rather than a color (Sol Ring produces `["C"]`), so a "count of colors" over it is ambiguous, and its data path (the C bit participates) differs from the two color fields.

Two tiny drive-by simplifications keep the touched functions inside the ruff complexity budget: the five identical tag-normalizer returns in `_rhs_to_json` collapsed into one table-driven return, and the twin color/identity explanation branches merged.

## Tests

- `api/parsing/tests/` — **1561 passed** (includes new parity + preprocess cases in `implicit_and_cases.py` and SQL-gen cases for `id>=3` / `id=2` / `id:2` / `id<2` / `id=0` / `c=2` / `c:3`)
- `api/tests/test_engine_unit.py` — **165 passed** end-to-end through the maturin-built engine, including new fixture tests (`c=2` → Boggart Ram-Gang + Kitchen Finks, `c>=3` → Monastery Messenger + Nicol Bolas, `id=0` → Black Lotus + Sol Ring, and `-id>=3` = exact complement 82/90)
- `cargo test` in `card_engine` — **157 passed** (new: an op×count popcount oracle over the plane fixture incl. the synthetic C-bit identity counting as zero, and JSON-builder cases incl. the `produced_mana` refusal)
- `cargo clippy --all-targets -- -D warnings` — clean
- ruff check + format — clean

Runtime validation against the full-corpus dev stack is pending (not run from this environment); the SQL was validated statically against the existing `color_identity_mask` tests and semantics.

### DIFF
--- a/api/db/2026-08-16-03-produced-mana-mask.sql
+++ b/api/db/2026-08-16-03-produced-mana-mask.sql
@@ -1,0 +1,33 @@
+-- produced_mana_mask(jsonb) encodes produced mana as a 6-bit integer (W=32, U=16, B=8, R=4, G=2, C=1).
+--
+-- SIX bits, where magic.color_identity_mask has five, and the extra one is not symmetry-breaking
+-- for its own sake: `produced_mana` is the one colour-ish column whose array can literally contain
+-- "C". Sol Ring's produced_mana is ["C"] where its colors and color_identity are both []. So a
+-- COUNT over this column counts colorless as a value, and Scryfall agrees — measured against
+-- api.scryfall.com 2026-08-16:
+--
+--   produces=6      = 106 = produces:all = produces:wubrgc   -- unreachable if C did not count
+--   produces=1 (C-only producers)             = 481          -- zero under a five-key count
+--   the three cards producing exactly {C,W} land in produces=2, and produces=1 has none of them
+--   counts 0..6 partition the corpus exactly: 30996+1143+504+147+10+693+106 = 33,599
+--
+-- The colour columns must keep counting FIVE and are deliberately NOT switched to this function:
+-- `c:all` = `c:wubrg` = `c=5` = 60 there, and `c=6` is not even a valid query ("Unknown color 6").
+--
+-- The expression index mirrors the color_identity one, so produces= / produces>= count queries are
+-- = ANY(mask_array) point lookups rather than a seq scan.
+
+CREATE OR REPLACE FUNCTION magic.produced_mana_mask(jsonb)
+RETURNS smallint LANGUAGE sql IMMUTABLE STRICT AS $$
+    SELECT (
+        CASE WHEN $1 ? 'W' THEN 32 ELSE 0 END +
+        CASE WHEN $1 ? 'U' THEN 16 ELSE 0 END +
+        CASE WHEN $1 ? 'B' THEN  8 ELSE 0 END +
+        CASE WHEN $1 ? 'R' THEN  4 ELSE 0 END +
+        CASE WHEN $1 ? 'G' THEN  2 ELSE 0 END +
+        CASE WHEN $1 ? 'C' THEN  1 ELSE 0 END
+    )::smallint
+$$;
+
+CREATE INDEX IF NOT EXISTS idx_cards_produced_mana_mask
+    ON magic.cards (magic.produced_mana_mask(produced_mana));
--- a/api/parsing/card_query_nodes.py
+++ b/api/parsing/card_query_nodes.py
@@ -213,9 +213,14 @@ def _compare_ints(lhs: int, operator: str, rhs: int) -> bool:
     raise ValueError(msg)
 
 
-def _color_count_masks(operator: str, count: int) -> list[int]:
-    """Bitmask values (5 colors => 2^5) whose popcount satisfies `popcount <op> count`."""
-    return [v for v in range(32) if _compare_ints(v.bit_count(), operator, count)]
+def _color_count_masks(operator: str, count: int, bits: int = 5) -> list[int]:
+    """Bitmask values whose popcount satisfies `popcount <op> count`.
+
+    `bits` is 5 for the colour columns and SIX for produced_mana, whose array can literally hold
+    "C" (Sol Ring produces ["C"] while its colors and color_identity are both []). See the
+    2026-08-16-03 migration for the measurements that pin the difference.
+    """
+    return [v for v in range(1 << bits) if _compare_ints(v.bit_count(), operator, count)]
 
 
 def _subset_masks(query_mask: int) -> list[int]:
@@ -532,7 +537,7 @@ class ExactNameNode(QueryNode):
 
 # What a colour-COUNT name (`c:m`, `id:gold`) means, per operator, as the (operator, count) pair
 # the numeric colour-count comparison is built from. Measured; the evidence and the two surprises
-# in it are written out at db_info.COLOR_COUNT_NAMES.
+# in it are written out at colors.COLOR_COUNT_NAMES.
 _COLOR_COUNT_BY_OPERATOR: typing.Final = {
     ":": (">=", 2),
     "=": (">=", 2),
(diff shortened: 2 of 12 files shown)


daveycodez:
### Routing: why this rode in on #879 rather than its own PR

This commit lowers a colour-COUNT *name* into the numeric colour-count node, so it cannot exist
without that node — and `FilterExpr::ColorCountCmp` / `_handle_color_count_comparison` are on
exactly one branch, this one. It was nearly routed onto **#926** (`parser-lang-operator`), which is
the natural *thematic* home since #926 is where the colour NAMES (`azorius`, `rainbow`,
`yore-tiller`) live; but #926 is based on `main` and lacks the node too — there, a numeric rhs on a
colour column falls through `_rhs_to_json` into `get_colors_comparison_object` and dies. It was
also nearly routed onto **#927** (`multilingual-store`), which turns out to carry neither the node
nor the colour names.

A stacked PR based on this branch is not expressible from a fork (GitHub requires the base branch
to exist in the target repo), so it lands here instead. **If #926 lands too, both touch
`_VALID_COLOR_NAMES` and `color_word` — a one-line merge each, union rather than replacement.**

---

### `c:m` is a count, and four of the seven operators do not mean what they say

`c:m` is Scryfall's word for MULTICOLOURED. It is not a colour — there is no colour `m` — and it spells no letters, so it cannot join `white` in `COLOR_NAME_TO_CODE`. It compares the NUMBER of colours in the column, which is exactly the comparison this branch already builds for `c>=2`. Six spellings mean it (`m`, `gold`, `multicolor`, `multicolour`, `multicolored`, `multicoloured`) and every one of them was a parse error.

**The operator table is measured, and it is not "substitute the number 2."** Corpus-wide against api.scryfall.com, 2026-08-16:

| query | total | equals | not |
|---|---|---|---|
| `c:m` = `c=m` = `c>m` = `c>=m` | 4,607 | `c>=2` | `c=2` is 3,811; `c>2` is 796 |
| `c<m` = `c!=m` | 29,049 | `c<2` | `c!=2` is 29,836 |
| `c<=m` | 33,599 | every card | `c<=2` is 32,812 |

`>` is the surprise on the high side — `c>m` is `c>=2`, not `c>2`, which those 796 three-and-more-colour cards separate — and `!=` is the surprise on the low side: `c!=m` is `c<2`, the negation of "is multicoloured", NOT `c!=2`, which the same 796 cards separate again.

`<=` is a **tautology** rather than `c<=2`, and it is pinned against a second term so it cannot be read as "the whole corpus": `c<=m t:creature` = `t:creature` = `c<=5 t:creature` = 18,753, where `c<=2 t:creature` = 18,140. It is spelled `>= 0` so it stays one leaf rather than becoming a node type.

The identity spellings take the same table on their own column: `id:m` = `id>=2` = 5,831; `id<m` = `id!=m` = `id<2` = 27,768 (`id!=2` is 28,824); `id<=m` = 33,599 = every card (`id<=2` is 32,543).

### Where the lowering lives

In `CardBinaryOperatorNode.__init__`, because **both parsers reach it** — the hand parser builds the node directly, the pyparsing one through `to_card_query_ast` — and a vocabulary added to one parser and not the other is precisely the failure `test_parser_parity` exists to catch. Everything downstream then sees the ordinary numeric node `c>=2` already produces: engine JSON, SQL and explanation alike. No new node type, no new engine variant, no schema change.

### produced_mana is deliberately left out

`produces:m` really is a count on Scryfall — 1,460, which is `produces>=2`. But it is a count over **six** values, colorless among them:

```
produces=1 produces:c   → 481    # exactly the cards that produce colorless and nothing else
produces>=1             → 2,603
produces:w or …:g       → 2,121  # the five-colour union — and this side agrees, at 2,126
```

That 482-card gap **is** the colorless producers. Both count implementations here read the five WUBRG keys (`magic.color_identity_mask` on the SQL path, the popcount in `ColorCountCmp` on the engine one), so a C-only producer would come back zero where Scryfall says one. Answering `produces:m` with the five-key count would be quietly short by 481 cards, so it stays the error it already was — the existing refusals in `_numeric_color_rhs_to_json`, `_handle_color_count_comparison` and `build_binary` now say *why*, and a test pins them. Making it work is a six-value count, which is its own change on both halves.

### Still invalid

`m` beside another colour letter stays a parse error, which is what Scryfall does with it too ("Using “m” with other colors is no longer supported" — the term is ignored): `c:mw`, `c:wm`, `c:mc`, `c:mm`, `c:mono`, `id:mw`, `produces:mw`.

### Tests

- 24 cases crossing each name against each operator, each asserting the value produces exactly the SQL its numeric comparison does, **in both parsers** (`test_color_count_names.py`), plus case, quoting and negation — 36 collected in all.
- The produced_mana refusals, by name and by bare number.
- The `m`-with-letters parse errors.
- Engine: the wire JSON the parser now emits for each lowering is walked to the row set it selects on the plane fixture, so a lowering that drifts shows up as a changed row set rather than as a filter that still builds.

`pytest api/parsing/tests` 1,736 passed / 19 xfailed. `cargo test --release -p card_engine` 159 passed. `cargo +1.97.1 clippy --all-targets -- -D warnings` clean. `ruff check` / `ruff format --check` clean.

daveycodez:
### produced_mana counts SIX values, because colorless is one of them

`produces:m` and `produces>=2` were refused on this branch, with the reasoning that produced_mana's
C key is "a genuine producible value, not a color, so a count over it would be ambiguous." The first
half is exactly right; the conclusion was wrong. It is not ambiguous — it is **six**.

Measured against api.scryfall.com, 2026-08-16:

| query | total | why it settles it |
|---|---|---|
| `produces=6` = `produces:all` = `produces:wubrgc` | 106 | a count no five-key popcount can reach |
| `produces=1 produces:c` | 481 | the C-only producers, counted as **one**, not zero |
| the 3 cards producing exactly `{C,W}` | — | answer `produces=2`; `produces=1` has none of them |
| counts 0..6 | — | partition the corpus exactly: 30,996 + 1,143 + 504 + 147 + 10 + 693 + 106 = **33,599** |

produced_mana is the one colour-ish column whose array can literally contain `"C"` — Sol Ring's
`produced_mana` is `["C"]` where its `colors` and `color_identity` are both `[]` — which is why it
counts differently from the other two rather than as an exception to them.

### The colour columns must keep counting five

That half is measured too: `c:all` = `c:wubrg` = `c=5` = 60, and `c=6` is not even a valid query
there (`Unknown color “6”`). Both widths are asserted **head-on**, not only through the popcount
oracle — an oracle that derives its expectation the same way the code does would happily follow a
later "cleanup" that unified the widths. The same C bit means *not a color* on an identity and *one
kind of mana* on the produced plane, and the tests say so in those words.

### Three places had to learn the width

- **`magic.produced_mana_mask`** — the six-bit twin of `magic.color_identity_mask`, with its own
  expression index so `produces=` / `produces>=` stay `= ANY(mask_array)` point lookups rather than
  seq scans.
- **`_color_count_masks(operator, count, bits=5)`** — takes a width, and the SQL builder picks the
  mask function by column.
- **`ColorCountCmp`'s eval arm** — masks `0b11_1111` for `ProducedMana` against `0b1_1111` for the
  other two.

### And then `produces:m` joins the colour-COUNT names

On its own operator table, which is the colour table intersected with "produces at least one value"
— a card that makes no mana is not a producer of anything, and Scryfall keeps it out of every one:

```
produces:m = produces=m = produces>m = produces>=m = 1,460 = produces>=2   (produces=2 is 504)
produces<m = produces!=m            = 1,143 = produces=1
    -- NOT produces<2 (32,139), which sweeps in the 30,996 cards producing nothing
produces<=m                         = 2,603 = produces>=1, not every card
```

The extra conjunct collapses into the count itself (`<2` ∧ `>=1` is `=1`; `>=0` ∧ `>=1` is `>=1`),
so this stays one (operator → operator, count) table and needs no AND node.

### Verification

`pytest api/parsing/tests` 1,740 passed / 19 xfailed. `cargo test --release -p card_engine` 158
passed. `cargo +1.97.1 clippy --all-targets -- -D warnings` clean. `ruff check` / `format --check`
clean. Against a local mirror the counts land on Scryfall's numbers exactly where the corpus
agrees: `produces>=2` 1,460, `produces=2` 504, `produces=6` 106, `produces=3` 147, `produces=4` 10,
`produces=5` 693 — and 0..6 partition our corpus too.

PR #889, opened 2026-08-07
Parsing: The Derivable is: Vocabulary — 43 Entries, Each Count- or Name-Validated

daveycodez:
## What

The rewrite-table half of the #872 split — every `is:` value whose membership derives from data the corpus already holds, following the 00713 discipline: **exact where achievable, documented divergence where coverage beats absence, never silent**.

- **Land cycles on `otag:cycle-*`** (per your review suggestion — it reached further than the three entries you commented on): fetchland, checkland, painland, slowland, bondland, battleland/tangoland, shockland, dual (ABUR), canopyland, scryland, fastland, triland, triome. Most name-verified against api.scryfall.com, not just count-matched — worth recording: a count-identical triland candidate held five wrong cards (SNC triomes instead of the Khans wedges), so name-level verification is part of the bar now.
- **`is:commander` refined per your comment**: `((t:legendary (toughness>=0 or t:background)) or o:"can be your commander") -banned:commander`. Adopted your toughness form (`>=0` because the parser rejects negative literals — `>-1` and `>=0` measure identical on Scryfall, 3720 both). Diffing the eligibility shape against Scryfall's `is:commander` surfaced something the docs don't say: **it excludes the commander banlist** (Griselbrand, Golos, Emrakul, Erayo were exactly the over-catch) while keeping 329 casual not-legal legends. Residual is your 00713 face-evaluation cluster: back-face legendaries over-match on combined type lines, and face-granted eligibility text under-matches until faces are searchable.
- **`companion`/`class`/`adventure`** from keyword/type/layout substrates (`is:adventure` is layout semantics by Scryfall's own definition — it equals `t:adventure or t:omen` there, 164 = 164).
- **Documented-delta entries** where the tag data is imperfect but coverage wins: bounceland (−2 untagged), filterland (−2), storageland (+5/−2), gainland (superset: with the refuge cycle the union covers every Scryfall member plus newer enters-tapped-gain-life cycles), frenchvanilla (community tag, ~+233 looser). Each delta name-diffed and recorded at its entry.

Not included, so the absences stay loud rather than approximate: shadowland/snarl (no tags exist), spell/modal (corpus-fuzzy), augment/host (never-legal-only cards), print-level values.

## Validation

- `api/parsing/tests/` + `client/tests/`: 1,726 passed (EQUIVALENCES extended for every entry; sampler synced)
- ruff + format clean
- Runtime-validated on a full-corpus dev stack: all cycle values exact, `is:commander` 3287, deltas as documented


api/parsing/rewrite.py
```diff
@@ -77,6 +77,70 @@
...
+    ("is", "commander"): '((t:legendary (toughness>=0 or t:background)) or o:"can be your commander") -banned:commander',
+    # Land cycles, rewritten onto the community cycle tags already present in
+    # Scryfall's oracle-tags bulk export (per review: membership as data the
+    # corpus carries, not an oracle-text signature that can over-match).
+    # Ancestor propagation makes the parent slugs work even though their
+    # taggings live on per-set children. Every mapping validated EXACT
+    # against api.scryfall.com's is: counts on 2026-08-07 (count in
+    # parens). Cycles whose tag taxonomy doesn't line up with Scryfall's
+    # is: membership (filterland tags 10 vs is: 22, gainland 25 vs 15,
+    # triome 5 vs 10) stay absent rather than approximate.
```
jbylund: You've got gainland/triome/filterland as not included here, but all 3 are included. Also if you wouldn't mind putting all the lands in a single contiguous alphabetized segment that would be great.

api/parsing/rewrite.py
```diff
@@ -77,6 +77,70 @@
...
+    # Scryfall member plus newer enters-tapped-gain-life cycles Scryfall's
+    # own list has not caught up with. frenchvanilla uses the community
+    # tag, whose definition runs looser than "keywords only" (~+233).
+    ("is", "bounceland"): "otag:cycle-rav-bounceland or otag:cycle-vis-karoo-land",
+    ("is", "filterland"): "otag:cycle-hybrid-filterland or otag:cycle-ody-filterland",
+    ("is", "storageland"): "otag:cycle-fem-storage-land or otag:cycle-mmq-storage-land or otag:cycle-tsp-storage-land",
+    (
+        "is",
+        "gainland",
+    ): "otag:cycle-ktk-gainland or otag:cycle-msh-gainland or otag:cycle-tmt-gainland or otag:cycle-zen-refugeland",
```
jbylund: How about `otag:gainland otag:cycle-dual-land` I think identical to what you have but doesn't require updating as new sets are updated? (as you say in your comment I think they're just behind on exactly this dimension).

api/parsing/rewrite.py
```diff
@@ -77,6 +77,70 @@
...
+    # spirit as is:bear above. bounceland and filterland are clean SUBSETS
+    # (-2 each: Arid Archway + Guildless Commons, and Cascading Cataracts +
+    # Crystal Quarry, are untagged upstream). storageland runs +5/-2 (the
+    # Mercadian depletion lands are tagged into the family; Crucible of the
+    # Spirit Dragon and Mage-Ring Network are not). gainland is a SUPERSET
+    # (+15, -0): with the refuge cycle included the union covers every
+    # Scryfall member plus newer enters-tapped-gain-life cycles Scryfall's
+    # own list has not caught up with. frenchvanilla uses the community
+    # tag, whose definition runs looser than "keywords only" (~+233).
+    ("is", "bounceland"): "otag:cycle-rav-bounceland or otag:cycle-vis-karoo-land",
```
jbylund: This is identical to `otag:bounceland` (no deviation + only one term / self updating).

api/parsing/rewrite.py
```diff
@@ -77,6 +77,70 @@
...
+    # is: membership (filterland tags 10 vs is: 22, gainland 25 vs 15,
+    # triome 5 vs 10) stay absent rather than approximate.
+    ("is", "fetchland"): "otag:cycle-fetchland",  # 10
+    ("is", "checkland"): "otag:cycle-checkland",  # 10
+    ("is", "painland"): "otag:cycle-painland",  # 10
+    ("is", "slowland"): "otag:cycle-slowland",  # 10
+    ("is", "bondland"): "otag:cycle-bondland",  # 10
+    ("is", "battleland"): "otag:cycle-tangoland",  # 10
+    ("is", "tangoland"): "otag:cycle-tangoland",  # 10; Scryfall accepts both names
+    ("is", "shockland"): "otag:cycle-rav-shockland",  # 10
```
jbylund: `otag:shockland` picks up https://scryfall.com/card/spm/180/multiversal-passage - but I think that fits the bill for shockland?

jbylund (commented):
I think my bias is a little towards using the generic tag rather than the union of specific tags - better chance of being self updating + the feedback the user sees

```
30 cards where (card oracle tags contains cycle-ktk-gainland or card oracle tags contains cycle-msh-gainland or card oracle tags contains...
```
vs
```
30 cards where card oracle tags contains gainland and card oracle tags contains cycle-dual-land
```

api/parsing/rewrite.py
daveycodez: Yea I would consider it a shock since you can pay 2, even though it's not one of the official dual shocks. I think just trusting the otag is good enough since that's community sentiment, pretty much across the board here. Getting exact match as scryfall isn't 100% necessary imo, but supporting these tags reaches more full syntax parity. My goal is to have full support of all scryfall syntax, even if the results might differ slightly

api/parsing/rewrite.py
daveycodez: Fixed — that paragraph was stale from before the documented-delta entries landed. All land entries (manland/creatureland included) now live in one alphabetized segment with a single policy comment.

api/parsing/rewrite.py
daveycodez: Swapped to otag:bounceland — and it's better than 'identical': the plain parent turns out EXACT at 17 (the two karoos my cycle-tag union missed, Arid Archway and Guildless Commons, were tagged on the parent all along). Single term, self-updating, no deviation.

api/parsing/rewrite.py
daveycodez: Took otag:gainland — 42 locally vs my union's 30, reaching the newer enters-tapped-gain-life cycles, and self-updating from here. (Read the second half of your comment as the dual-land parent for context rather than a proposed change to is:dual — kept that one on cycle-abu-dual-land since the plain parent aggregates every dual cycle via propagation, ~423, which is much broader than Scryfall's ABUR-only meaning. Say the word if you'd rather is:dual go wide too.)

api/parsing/rewrite.py
daveycodez: Swapped to otag:shockland (11, Multiversal Passage included) — and as I said above, that's the policy for these now: trust the community tag, deviations from Scryfall's exact membership are acceptable, syntax coverage is the goal.

api/parsing/rewrite.py
```diff
@@ -71,12 +71,82 @@
...
+    ("is", "frenchvanilla"): "otag:french-vanilla",  # community tag, ~+233 looser than "keywords only"
+    # is:modal by the mode-introducing wording. Runs ~+50 over Scryfall's own
+    # count (835 vs 800 live / 770 paper on 2026-08-09): the union reads any
+    # "choose ..." mode header, which is broader than their curated notion --
+    # accepted as a documented delta, per the trust-the-wording policy the
+    # land segment already follows.
+    (
+        "is",
+        "modal",
+    ): 'o:"choose one" or o:"choose two" or o:"choose three" or o:"choose four" or o:"choose up to"',
```
jbylund, line 149: This is over by cards like https://scryfall.com/card/znr/89/acquisitions-expert , there's a modal oracle tag which agrees almost perfectly.

PR #912, opened 2026-08-09
API: Every Scryfall /cards/* Route, So the Base URL Is the Only Thing a Client Changes

daveycodez:
> [!IMPORTANT]
> **Stacked. Merge order: #894 → #913 → this.**
> It builds on the pristine `raw_card_blob` from #894 and the `CardOrdering` / `SortDirection.AUTO`
> members from #913, so the diff here includes both until they land. Nothing else to do afterwards —
> there is no follow-up PR.

## Motivation

The engine already answers Scryfall's query syntax, and `/search` returns it in this project's own response shape — selected fields, a `cards` array, `limit`/`offset` — which is what the frontend wants and is untouched here.

A Scryfall-shaped client wants something different: Scryfall's response objects and pagination, plus the per-card addressings it reaches for alongside search — `/cards/:id`, `/cards/named`, `/cards/:code/:number`, the external id namespaces, rulings. This adds that as a second surface, so such a client can be pointed here by changing its base URL and nothing else.

Doing the whole surface at once is why #878 (`/cards/named` alone) was closed in favour of this; its fuzzy design, thresholds included, is carried over here.

## What this adds

| Route | Parameters |
| --- | --- |
| `GET /cards` | `page` |
| `GET /cards/search` | `q`, `unique`, `order`, `dir`, `page`, `format`, `pretty`, `include_extras`, `include_multilingual`, `include_variations` |
| `GET /cards/named` | `exact` / `fuzzy`, `set`, `format`, `face`, `version`, `pretty` |
| `GET /cards/autocomplete` | `q`, `include_extras` |
| `GET /cards/random` | `q`, `format`, `face`, `version` |
| `POST /cards/collection` | ≤75 identifiers, `not_found` echoed back |
| `GET /cards/:id` | Scryfall UUID |
| `GET /cards/:code/:number(/:lang)` | language defaults to `en`, as Scryfall's does |
| `GET /cards/multiverse\|mtgo\|arena\|tcgplayer\|cardmarket/:id` | MTGO also matches `mtgo_foil_id`, TCGplayer also `tcgplayer_etched_id` |
| `GET /cards/:id/rulings` + four sibling addressings | |

Scryfall's List / Catalog / Ruling / error objects, its 175-per-page pagination with a followable `next_page`, `format=text` and `format=image` (a 302 to the card's own CDN URL — the same place Scryfall sends you), `format=csv`, `pretty=`.

**No routing change was needed.** `_resolve_action` matches a full path before falling back to the first segment, so the five named sub-routes claim their exact paths and the other nine shapes arrive at one handler as ≤3 positional segments. The handlers are a mixin on `APIResource` — `iter_marked_routes` scans inherited attributes, so they register like any other route while the compatibility surface stays in `api/scryfall_compat/`.

## Two things the stored data did not support

**The card object itself.** `raw_card_blob` is snapshotted *after* the importer adds `card_name`/`face_name`/`face_idx` and normalizes absent `flavor_text` to `""`. Both are exactly reversible and `to_scryfall_card` reverses them — `""` is a value Scryfall never sends, so absent and empty are one state.

This nearly needed a second column. The blob for a multi-face row used to be **the front face promoted to top level** — `card_faces` gone, `name` and `type_line` the front's, eleven extra keys — and no reader can undo that, because which keys a real card carries at top level depends on layout (a split card has `mana_cost` and `image_uris` there, a transform card does not). This PR originally carried a `cards.scryfall_json` column holding the untouched object for those rows. That was deleted in favour of fixing the blob at the source in #894, which this PR is now stacked on: exactly one thing depended on the promotion, `image_uris`, and every reader of it already coalesced to `card_faces->0`.

**Deployment note:** rows are rewritten by an import, not by a migration, so until the first bulk import after deploy a multi-face row still holds a promoted front face and is served as its card rather than as a `card_face`. One import cycle, ~2% of rows, closes on its own.

**Rulings did not exist.** New `magic.rulings`, loaded from the bulk `rulings` file the fetcher already knew about, wired in after the tag imports. A whole-table replace in one transaction rather than an upsert — the file carries no ruling id, and rulings are occasionally retracted, so insert-only would accumulate withdrawn rows. `DELETE` rather than `TRUNCATE` so readers keep seeing the previous contents through MVCC, and a failure is logged rather than raised: rulings are the only thing in the import sequence nothing else reads.

## Indexes, verified on the 97,808-card corpus

The ids Scryfall routes by live inside `raw_card_blob`. Each partial index predicates on the *same* expression its lookup compares, so the planner can prove the query implies the predicate.

| Lookup | Plan |
| --- | --- |
| `/cards/mtgo/:id` | Index Scan, `idx_cards_mtgo_id` |
| `/cards/multiverse/:id` | Bitmap Index Scan, `idx_cards_multiverse_ids` (GIN `jsonb_path_ops`) |
| `/cards/:code/:number/:lang` | Index Scan, `idx_cards_set_collector_lang` |
| `/cards/named?exact=` | BitmapOr across three folded-name indexes |
| `/cards/named?fuzzy=` trigram stage | 5.6 ms |

Two of these were found by measuring, not by reading: indexing `card_set_code` rather than `lower(card_set_code)` cost the index entirely (the planner fell back to `idx_cards_collector_number` and rechecked set and language per row), and `named?exact=` also matches one face of a `Front // Back` card, which is why two `split_part` expression indexes sit alongside the whole-name one.

## `/cards/random` must not be cached, twice over

The draw is non-deterministic for a fixed query and parameter set, which is exactly what both cache layers assume away. `_run_query` keys on SQL text plus parameters — both invariant here — so it would have replayed the first card forever; the draw goes through `_run_uncached` instead. `CachingMiddleware` would have pinned one card per import generation; the response sets `Cache-Control: no-store`.

Worth flagging for review: the test that catches this **passes on a buggy implementation under default suite settings**, because with caching off `_query_cache` holds one entry and the count query and the draw evict each other. The test installs the multi-entry cache production has before drawing.

## Known divergences

Full list in the issue doc; the ones that matter:

1. **The corpus is a filtered subset.** Cards illegal in every format, tokens, playtest cards, funny sets and digital-only printings are never imported, so a card this instance does not hold 404s here and resolves on Scryfall. Project policy, not a property of these routes, and the largest single difference.
2. **Query syntax parity is tracked separately** — `q` goes to the existing parser, so the in-flight parser work applies here automatically.
3. **Two `order` values are unsupported** — `penny` and `review`. They fall back to `name`, which is what Scryfall itself does with an order it does not recognize (measured: it falls back silently rather than erroring), plus a `warnings` entry saying so. The other six of Scryfall's vocabulary are wired by #913, and `_ORDER_MAP` here is built from `CardOrdering`, so nothing needs adding twice.
4. `dir=auto` resolves to ascending; the three `include_*` flags are accepted and ignored (there are no extras in the corpus to include).
5. `format=csv` column parity and `format=text` rendering are faithful but not verified cell-for-cell against Scryfall.
6. Card URIs are **not** rewritten — `uri`/`rulings_uri`/`prints_search_uri` still point at Scryfall, because they are part of the payload. `next_page` is the exception, and takes its scheme from `Forwarded`/`X-Forwarded-Proto`. Guessing `https` from the hostname was tried and rejected: it breaks plain-HTTP deployments on real hostnames to paper over a misconfigured proxy.

## Merge safety

**Merge order is #894 → #913 → this.**

- **#894** — this serves the card object straight out of `raw_card_blob`, which is only the card Scryfall sent as of #894's last commit.
- **#913** — `_ORDER_MAP` is built from `CardOrdering`, and `dir=auto` forwards `SortDirection.AUTO` to `_search` rather than resolving it here, so both endpoints agree on what auto means. Those members are #913's.

Also verified to merge cleanly against **apply-query-directives (#893), response-fields, engine-regex-parity and numeric-identity-compares**, with the suite re-run green after merging #893.

## Validation

- 114 new tests: `test_scryfall_compat_objects.py` (pure payload construction) and `test_scryfall_cards_routes.py` (every route through `_handle`, plus a class that goes through a real `falcon.App` — media-handler resolution against the `charset=utf-8` content type, the 302, HEAD, a posted collection body, and following `next_page` to walk a result set).
- Full suites green: `api/ card_engine/ scripts/ client/` → **2,738 passed**, 19 xfailed; `cargo test --release` 156 passed; jest 1,747 passed.
- `ruff check` and `ruff format` clean on the touched files.
- Migration and index plans exercised against the real dev corpus, not just the test fixtures.


daveycodez:
### Merge order for the five PRs that carry the card-object and by-name code

These five all contain copies of the same files at **different generations**, so the order they land in decides whether a merge moves the code forward or silently backward. Measured at today's heads — #912 `36e2d5b3`, #922 `80ff051c`, #927 `209fb11f`, #928 `a7059beb`, #929 `5fd256c9`:

| PR | `card_engine/src/card_object.rs` | `card_object_parity.json` | engine by-name (`name_best`) | `ARCHIVE_FORMAT_VERSION` |
|---|---|---|---|---|
| #912 | newest — marketplace search fallbacks | present | yes | **2026083101** |
| #922 | same as #912 (**contains** it) | present | yes | 2026082703 — behind #912 now |
| #927 | older — ids-only `purchase_uris`, no search fallbacks | absent | no (SQL only) | 2026082707 |
| #928 | older | absent | yes | 2026082707 |
| #929 | older | absent | no (SQL only) | 2026082708 |

**Correction to the first version of this comment:** it opened with "#912 first", which is only true *within these five*. The PR descriptions already carry the full chain and it starts earlier — **#894 → #913 → #912 → #893 → #927 → #928 / #929**, with #922 any time after #912. Nothing below changes that order; this comment is about the conflicts it produces, not a replacement for it.

**Within these five: #912 → #922 → #927 → #928 → #929.**

- **#912 before the other four** because it owns the newest copies. Land it first and every later branch resolves *toward* `main`; land it last and each earlier merge drags `main` backward and #912 arrives as a giant conflict. (#894 and #913 still come before it, per its description.)
- **#922 second, and not before #912** — it already contains #912's commits, so merging it first lands #912's whole feature under #922's review.
- #928 and #929 each contain an **older** #927; #927 has since moved ahead, so whichever of the three lands later must take #927's newer commits rather than its own copy.

#### Resolution rules for the conflicts this will produce

1. **`card_object.rs` / `objects.py` / `purchase_uris`: `main`'s copy wins after #912 lands.** Re-apply only the branch's own additions on top. Taking the branch's older copy wholesale reverts the eleven card-object fixes and the marketplace search fallbacks, and reintroduces `Snake // Zombie` being searched for as `Snake` (api.scryfall.com searches the joined name on `split`, `reversible_card` and `double_faced_token`, the front face otherwise).
2. **`api/scryfall_compat/fixtures/card_object_parity.json` is #912-only.** Expect modify/delete conflicts on the other branches — **keep the file**. Never regenerate it to resolve one: regeneration reorders every key and buries the real diff, which is six URL lines across two cases (`Fire // Ice`, `Propaganda // Propaganda`).
3. **`ARCHIVE_FORMAT_VERSION`: the HIGHEST value must win**, not whichever side git picks. **This changed after the first version of this comment** — #912 now carries **2026083101**, above #929's 2026082708, #927/#928's 2026082707 and #922's 2026082703, because its `card_name_lower_id` fix is itself a layout change and is dated today so it sits above all four whichever order they land. So the value to keep is #912's, and a merge that quietly takes a lower one leaves the merged code building stores that claim a format they do not have.
4. **The name-rule SQL fragments are identical on purpose.** `_collate_name`, `_collated_sql`, `_NAME_SPLITS_IN_TWO`, `_FACE_NAME_MATCH`, `_WHOLE_NAME_MATCH`, `_COLLECTION_NAME_MATCH`, `_EXACT_NAME_MATCH` and `_WHOLE_NAME_FIRST` were copied character for character into #927/#928/#929 so no two branches can answer the same needle differently. If a conflict appears there, the correct resolution is "they are the same text" — confirm, don't rewrite one of them.
5. **One cleanup that only becomes possible after #912 and #928 are both in.** #912's `name_best` runs a second, full-scan pass on a miss because *its* `name_trigram` index is built over the folded name, so a collated needle (`limduls vault`) shares no trigram with the stored spelling. #928 builds that index over the **collated** name, which makes the narrowed pass sound on its own and the second pass dead weight. Remove it deliberately once both have landed — not as part of a conflict resolution.

daveycodez:
Heads up on a security issue in this PR, plus a fix I've pushed to the branch.

`_self_base_url` builds `next_page` from `request_host`, which dispatch fills from an unvalidated `X-Proxy-Host`, and takes its scheme from `request.forwarded_scheme` (`Forwarded` / `X-Forwarded-Proto`, also unvalidated). This PR is also what adds `public, max-age=57600` to `/cards/*` — its own comment notes those routes previously sent no `Cache-Control`, so a CDN cached none of them.

Those two halves together are a web cache poisoning vector. One request carrying `X-Proxy-Host: evil.example` seeds a 16-hour cached response, keyed by a URL anyone can request, whose `next_page` points off-host. `X-Forwarded-Proto: http` is the same defect as a scheme downgrade. Neither header is part of a URL-based cache key.

**This is latent, not live.** I checked sylvan-librarian.com and `/cards/*` returns 404 there today, and the routes listing carries no `/cards/*` surface, so nothing on the deployed site currently builds a URL from these headers. It arms on merge. Card objects are unaffected either way: every `to_scryfall_card` call uses the default `https://api.scryfall.com` base, so `uri`, `set_uri`, `rulings_uri` and `prints_search_uri` were never reachable. Only `next_page` was.

The fix honours `X-Proxy-Host` only from a `TRUSTED_PROXY_HOSTS` allowlist — exact, whole-string, case-insensitive, no suffix matching, so a rule for `cards.example.com` can't admit `evil-cards.example.com`. The forwarded scheme rides on an allowlisted host and is clamped to `http`/`https`. Unset is the default and ignores both headers, building from the request's own host and scheme, which is the correct answer for a deployment fronted directly. The site-name path is untouched.

178 tests in the module pass and ruff is clean. I rewrote `TestSelfUrlScheme`, which had been asserting the vulnerable behaviour.

One thing I can't determine from outside, and you can: **your nginx cache key for `/cards/*` once these routes go live.** The homepage is keyed by full URL, and query-string URLs appear to bypass the cache entirely. If `/cards/*` inherits that behaviour, the poisoning is real post-merge; if it's keyed or bypassed differently there, the exposure changes. Worth confirming before this lands.

daveycodez:
Added `e1d16862`: `is:atypical` / `is:default` as an engine leaf. Both were the two largest `is:` values still answering zero (10,423 / 33,267 on api.scryfall.com, 2026-09-03). `FilterExpr::Atypical` calls the same `printing_is_atypical` over the same `PreferClassIds` that `prefer:atypical` ranks by, so the filter and the prefer cannot disagree about which printings are the class; `default` is its `Not`, which is what `is:atypical is:default` = 0 and `is:default -is:atypical` = 33,267 say it is. No parser change on main (the words arrive as ordinary `is:` tags and the engine claims them); #926 lists them in `ENGINE_IS_VALUES` so its warning gate stays quiet. Clippy 1.97.1 clean, 220 engine tests, 189 engine unit tests.

PR #915, opened 2026-08-10
Search: Stop a Bare Mana Value Dropping Symbols It Does Not Recognise

jbylund:
Stacked on #909 — review that first; this PR's diff is only the three commits' worth of mana work on top.

## The bug

`mana:s` matched 31,726 cards. `mana:{s}` matched 3, which is the right answer.

`mana_cost_str_to_dict` read the unbraced form against its own hardcoded charset, `"WUBRGCX"`, and silently dropped anything absent from it. `S` and `P` were absent, so a bare snow or phyrexian pip vanished and left an empty cost dict — and an empty JSONB object is a subset of every cost, so `{} <@ mana_cost_jsonb AND cmc >= 0` matched every card with a mana cost.

Dropping quietly is the worse half. `mana:snow` kept only the `W` and generated SQL byte-identical to `mana:w` — 6,379 cards, none of them an answer to the question asked. That is a *wrong* result, not a broad one.

## Root cause

Five copies of "what is a cost symbol", each missing something different:

| location | charset | missing |
|---|---|---|
| `card_query_nodes.mana_cost_str_to_dict` | `WUBRGCX` | S P Y Z |
| `card_query_nodes.calculate_cmc` | `\d+\|[WUBRGC]` | S P X Y Z |
| `pyparsing_based.simple_mana_symbol` (×2) | `0-9WUBRGCXYZ` | S P H |
| `mana_symbols.MANA_COST_ATOMS` | authoritative | — |

All now read `MANA_COST_ATOMS`, and the bare form is *validated* rather than filtered, so an unrecognised character is refused instead of ignored.

## Verification

Counts from the generated SQL run against a full card table:

```
mana:s        31726 -> 3       == mana:{s}
mana:p        31726 -> 0       == mana:{p}
mana:snow      6379 -> 400     (had been answering mana:w)
mana:hello    31726 -> 400
mana:w        17980 -> 17980   == mana:{w}       unchanged
mana:2ww               2821    == mana:{2}{w}{w}
mana:x                 1991    == mana:{x}
```

Every bare form now agrees with its braced equivalent. Suites: 2054 passed / 21 xfailed, jest 1834 passed, ruff + prettier clean.

## Stored data is unaffected

`mana_cost_str_to_dict` also builds `mana_cost_jsonb` at ingestion, but **0 of 97,808** cards have an unbraced cost, so ingestion has always taken the braced branch. Query-side only — no reindex.

## Parser divergence

Fixing pyparsing's charset settles four of the six ways the two parsers disagreed here (`mana:s`, `mana:p`, `mana:snow`, `mana:red`). The remaining two — `mana:hello`, `mana:hw` — share one cause: a MANA field falls back to pyparsing's generic string value when its mana pattern misses, which the hand parser has no equivalent of. Fixing that means making the mana field's value parser mana-only, a grammar change rather than a widened character class, so it is recorded in `KNOWN_DIVERGENCES` instead.

That entry did not come from the wild-query corpus. It holds no `mana:` or `devotion:` query at all, which is why no sweep had found any of this.


jbylund:
Closing rather than rebasing — conflicts are substantive, not just mechanical. This branch predates #941 (merged via #909), and this PR's `MANA_COST_ATOMS` constant still treats bare `P`, `Y`, `Z` as valid atoms; #941 established none of those three ever appears unpaired in a real cost, so reusing this constant as-is would regress that fix.

The underlying bug this PR diagnosed is still fully live on main today (verified: `mana:s`/`mana:snow`/`mana:p`/`mana:hello` all still silently produce wrong or empty results via the pyparsing path). Refiled as #954, which credits this PR's diagnosis and flags the P/Y/Z conflict for whoever picks it up next.

PR #922, opened 2026-08-12
API: The Rest of the Scryfall Surface — /sets, /catalog/* and /symbology

daveycodez:
> [!IMPORTANT]
> Stacked on **#912**. Merge #912 first. GitHub will not let this PR use `scryfall-cards-api` as its
> base — that branch lives in the fork and a base branch must exist in this repository — so the diff
> below shows #912's commits until it lands. The commits that belong to this PR are the last three.

## Motivation

#912 made every route under `/cards` answer the way api.scryfall.com does. That left the half of the
API that describes Magic rather than returning cards: what sets exist, what the game's vocabularies
are, what the mana symbols mean. A client pointed at this host got its cards and then 404d on
`/sets`. This adds the remaining twenty-six routes, after which the surface is complete.

## What

| Route | Answers with |
| --- | --- |
| `GET /sets` | List of 1,047 Set objects |
| `GET /sets/:code`, `/sets/:id` | One Set object |
| `GET /sets/tcgplayer/:id` | One Set object, by TCGplayer group id |
| `GET /catalog/:name` | Catalog object, over the twenty documented names |
| `GET /symbology` | List of 84 CardSymbol objects |
| `GET /symbology/parse-mana?cost=` | ManaCost object, computed rather than stored |

Three new tables, an importer wired into the existing sequence beside the rulings load, and a second
routes mixin.

## Mirrored, not derived — **the part worth reviewing**

Everything except `parse-mana` is served from a table mirrored off Scryfall rather than computed from
`magic.cards`. That is not the obvious choice — this project already derives `/cards/autocomplete`
from its own corpus — so the reasoning matters:

- **A Set object carries eight fields no card carries**: `tcgplayer_id`, `mtgo_code`, `arena_code`,
  `icon_svg_uri`, `block`, `block_code`, `parent_set_code`, `printed_size`. `/sets/tcgplayer/:id` is
  not degraded without the first of them, it is unimplementable from the corpus — a set's
  `tcgplayer_id` is a TCGplayer **groupId** (`24766` for `trk`) while a card's is a **productId**
  (`706132`–`706191` for cards in that same set). Different namespaces, no derivation between them.
- **`card_count` counts Scryfall's printings.** This corpus is a deliberate subset, so a derived
  count would report a number no other Scryfall client agrees with.
- **A card symbol has no card** to derive `svg_uri` or `gatherer_alternates` from.

## parse-mana, and two rules nothing documents

Both measured against the live API rather than inferred:

1. **The normalized cost reorders colored pips into canonical colour order** — `RUW` answers
   `{U}{R}{W}`. Every canonical ordering (allied pairs, enemy pairs, shards, wedges, four-colour
   runs, WUBRG) is a walk around the colour wheel at a constant step: one for anything contiguous,
   two for anything not. Step 1 before step 2, starting points in WUBRG order, reproduces Scryfall
   for **all 31 colour combinations**.
2. **Emission order is X, generic, colored pips, `{C}`** regardless of input order, generic summed —
   `2XWU` answers `{X}{2}{W}{U}`, `1{1}` answers `{2}`.

Also: an empty cost is `null` but `cost=0` is `"{0}"`, and an unparseable fragment is a **422**.

These are exactly what a hand-written test would re-assert from the implementation, so
`test_scryfall_mana.py` pins **79 goldens captured from the live API** instead, covering all 31
colour subsets written forwards and backwards.

## Cache headers are mirrored too

These routes do **not** carry the card routes' `public, max-age=57600`:

| Route | Cache-Control |
| --- | --- |
| `/sets`, `/sets/:code`, `/sets/tcgplayer/:id`, `/catalog/*`, `/symbology` | `public` |
| `/symbology/parse-mana` | `max-age=0, private, must-revalidate` |

The first cut of this branch applied the card tier to all six, which meant telling shared caches to
hold `parse-mana` for sixteen hours where Scryfall revalidates. Both upstream values are mildly
surprising and are mirrored rather than endorsed — reasoning in the design doc. An A/B against the
live API returns 4/4 byte-identical.

## Validation

Full suite **2,970 passed, 19 xfailed** (up 142 from 2,828 on #912's head); `ruff check` clean and
`ruff format --check` clean across every `.py` in `upstream/main...HEAD`.

End to end against live Scryfall on 2026-08-11, in a fresh container:

| Step | Result |
| --- | --- |
| `import_sets` | 1,047 sets in 0.6 s, served order matching upstream **position for position** |
| Set addressings | all four resolve the same set by code, id and TCGplayer id |
| `import_catalogs` | 20 catalogs, 62,187 values in 2.65 s, zero failures, each served back at the count imported |
| `import_symbology` | 84 symbols in 0.1 s |
| `parse-mana` | matches upstream on every probe |
| Cache headers | 4/4 byte-identical to upstream |
| Re-import | stable at 1,047 — the replace is idempotent |

Worth flagging: the first cut of the `_import_reference_quietly` test stubbed only the failing step,
leaving the other two to perform **real Scryfall fetches and real writes into the shared test
database** — while passing. The helper now stubs all three and says why.

## Known divergences

1. `card_count` is Scryfall's number, not this instance's.
2. `/catalog/card-names` and `/catalog/artist-names` list cards and artists this instance may not
   hold, for the same corpus reason #912 already owns.
3. A catalog name Scryfall adds later 404s until the list in code is extended — deliberate, so an
   unknown name is not served as a silently empty catalog.
4. `uri` / `search_uri` / `icon_svg_uri` on a Set object point at Scryfall, matching #912's rule that
   payload URIs are not rewritten.

Full reasoning and follow-ups in
[docs/issues/local-scryfall-sets-catalogs-symbology.md](https://github.com/daveycodez/sylvan_librarian/blob/scryfall-sets-catalogs-symbology/docs/issues/local-scryfall-sets-catalogs-symbology.md).

## Merge safety

The migration is additive — three new tables, no change to any existing one — and sorts after
`2026-08-10-01`. Nothing outside the new mixin reads the new tables, and the only edits to #912's
files are the `ScryfallResponder` extraction (two methods moved, no behaviour change), the mixin
being added to `APIResource`'s bases, and one `ruff format` blank line.

---

## Follow-up (2026-08-16): six answers measured against api.scryfall.com and corrected

This surface was written from Scryfall's documentation and from what looked reasonable. Put to
api.scryfall.com one request at a time, six of its answers were not Scryfall's. Every change below
is a measurement, and the commit messages carry the request/response pairs.

| Behaviour | This branch answered | api.scryfall.com answers |
|---|---|---|
| `/symbology/parse-mana` with no `cost` | `400 "You must provide a cost parameter to parse."` | `200 {"cost": null, ...}` — the same body `?cost=` gives |
| `?cost={W/U/B}` | `200`, a three-coloured ManaCost | `422` — a hybrid has exactly two halves |
| `?cost=xyzzy` | `{X}{Y}{Z}{Z}{Y}` (writing order) | `{X}{Y}{Y}{Z}{Z}` — sorted and grouped |
| `?cost=!!!` | `“{!}”` — the first fragment, re-braced | `“!!!”` — every fragment, as written |
| `/catalog/Card-Types` | `200` | `404` — catalog names are case-sensitive |
| A 404 about the PATH | `public` | `no-cache` (a 404 about the DATA, `/sets/zzzz`, keeps `public`) |

Plus two that are not per-route: `/sets/:code/extra` answered "No Magic set found for the given code
or ID" about a set that was fine, and **every error body is indented** on api.scryfall.com — measured
across the whole surface, and it does not negotiate (`Accept: application/json`, `Accept: text/html`,
a bare wildcard and an explicit `?pretty=false` all give the same 130-byte indented not-found, while
every data body is compact). This rendered both compact, so a client comparing bytes saw a different
document for every 4xx it received.

The unparseable-fragment wording turned out to follow one rule rather than five, and the six new
goldens pin it: what comes back is the input with everything Scryfall could read struck out, adjacent
bare characters merged into one run, and a braced token keeping its braces. `{W/U/B}` answering
`“{//}”` is what made the rule visible.

**One measurement is recorded rather than reproduced.** A 133-character nonsense cost comes back
naming 51 characters. One data point does not determine a truncation rule, so nothing was guessed.

Two existing tests asserted the behaviour Scryfall does not have (`test_the_name_is_case_insensitive`,
`test_a_missing_cost_is_a_400`) and are inverted rather than deleted, since the URLs they cover still
need covering.

Found by a differential sweep of the peripheral API surface — response formats, the reference
endpoints and HTTP-level behaviour — against api.scryfall.com, run from the Cloudflare port
(daveycodez/sylvan-librarian-cloudflare, `scripts/parity-sweep.ts`). That harness now carries 200
cases over this surface, so these stay covered.


daveycodez:
### Merge order: this one is **second**, and must not land before #912

Full ordering and conflict-resolution rules are in https://github.com/jbylund/sylvan_librarian/pull/912#issuecomment-5487092699 — the short version for this PR:

This branch **contains #912's commits** (it is the only true ancestor relation among the five: `git merge-base --is-ancestor origin/scryfall-cards-api origin/scryfall-sets-catalogs-symbology` is true at `80ff051c`). Merging it first would land #912's entire card-object and by-name feature under this PR's review. Take #912 first; this PR's remaining diff is then just `/sets`, `/catalog/*` and `/symbology`.

It carries the same generation as #912 (search fallbacks present, `card_object_parity.json` present, `ARCHIVE_FORMAT_VERSION` 2026082703), so once #912 is in there should be nothing to resolve here beyond its own feature.

PR #926, opened 2026-08-16
Parsing: lang:, oracleid:, the Colour Names, and the Quotes Every Keyboard Produces

daveycodez:
Based on `main` (5b2c52a) — independent of every other PR in this series; merge first, it
unblocks parser-fixture regeneration downstream.

Two operators, one class of change: both are new `FieldInfo` entries in `DB_COLUMNS` joined to
the exact-match family in `card_query_nodes.py`, and both are Scryfall operators this parser
rejects today.

### `lang:`

`lang:` (alias `language:`) as a search operator: a new `card_lang` FieldInfo in `DB_COLUMNS`
(`FieldType.TEXT`, `ParserClass.TEXT`), joining the exact-match family `card_set_code` /
`card_layout` / `card_border` / `card_watermark` — the value is lowercased at parse time so
`lang:JA` and `lang:ja` are the same query, and `:` means equality, not pattern matching.
`CardAttributeNode` gains the "language" human-explanation label.

Scryfall grammar notes, verified against the live API:

- `lang:ja`, `lang:any` parse; `lang:any` is the "every language" widener (its meaning is enforced
  by the engine/SQL consumers, not the parser).
- `-lang:en` negates like any other text equality.
- `lang>=en` errors exactly like string-order comparisons on the other string columns — the
  operator grants equality only.

### `oracleid:`

`oracleid:` (alias `oracle_id:`) mapped to the existing `magic.cards.oracle_id` column, in the
same exact-match family, with the "oracle ID" human-explanation label.

**This one closes a self-inconsistency in our own API.** `api/scryfall_compat/objects.py` builds
`prints_search_uri` for every card object as

```
{base_url}/cards/search?order=released&q=oracleid%3A<oracle_id>&unique=prints
```

(`objects.py:307`, on the `scryfall-cards-api` branch). Every card object we emit hands the client
a search URI that our own `/cards/search` answers with a parse error, because `oracleid` is in no
`FieldInfo`'s `search_aliases`. The same URI is emitted by the Cloudflare deployment's TypeScript
port of that serializer, where it was caught by a live-parity run against api.scryfall.com and
recorded as a known deviation. Adding the operator makes the link resolve.

`oracle_id` is the one searchable UUID column and every parameter this parser binds is text, so
`CardAttributeNode.to_sql` renders it as `card.oracle_id::text`. Postgres has no `uuid = text`
operator, and comparing the column's canonical text form also means an unparseable search value
matches nothing rather than raising mid-query. The search value is lowercased for the same
reason `set:`/`lang:` are — a uuid renders lowercase through `::text`, and Scryfall's `oracleid:`
is case-insensitive.

Scryfall grammar notes, verified against the live API (2026-08-16):

- `oracleid:<uuid>`, `oracle_id:<uuid>`, `oracle-id:`, `oracleId:` all resolve — Scryfall folds
  case and punctuation in keywords. `oid:` and `oracleids:` are rejected as unknown keywords, so
  the alias list stops at the two spellings this parser's alias table can express.
- The value is a canonical dashed v4 uuid, case-insensitive, quoted or bare. `unique=prints`
  returns every printing sharing the id (4 for Jötun Grunt); default `unique=cards` rolls up to 1.
- A well-formed uuid naming no card is an ordinary empty result (404 `not_found`).
- A malformed uuid is **not** a parse error on Scryfall's side either: the term is dropped with a
  `warnings` entry, and the request 400s only when nothing else survives — `oracleid:not-a-uuid
  t:creature set:cmd` returns 200 with 135 cards and the warning attached. That ignored-terms
  mechanism does not exist in this parser, so validation stays out of it entirely, exactly as
  `lang:` treats unknown language codes: a malformed id parses as an ordinary string that no
  stored id equals.
- Ordered comparisons parse and match nothing on Scryfall too; here they parse the same way they
  do for the other string-equality columns.

## Why it lands first and alone

The parser is ported (TypeScript) in the Cloudflare deployment, and its parity fixtures are
regenerated from THIS Python parser — a port-side `lang:` or `oracleid:` with no upstream twin
fails the parity gate by construction. This PR is deliberately just the operators: for `lang:`,
the column it compares against (`magic.cards.card_lang`) and the engine's `LangMatch` widening
ship with the multilingual-store PR, which follows. `oracleid:` needs nothing further on the SQL
side — `magic.cards.oracle_id` already exists and is `NOT NULL`.

## Validation

- `api/parsing/tests/test_lang_parsing.py` — equality both quoted and bare, `lang:any`,
  negation, unknown codes (parse fine, match nothing — Scryfall's behavior), the `lang>=` error
  parity case, alias spellings, and non-ASCII quoted values alongside.
- `api/parsing/tests/test_oracleid_parsing.py` — both alias spellings, quoted and bare, uppercase
  uuid folding, negation, ordered comparisons, boolean combinations, a well-formed id that names
  no card, a malformed id, and the `::text` cast pinned so a future refactor cannot reintroduce
  `uuid = text`.
- Both files run against **both** parser implementations (hand-rolled and pyparsing) through the
  shared `parse_query` fixture.
- Full parser suite: **1,764 passed, 19 xfailed** (the pinned #904 divergences, unchanged).
- `ruff check` / `ruff format --check` clean.
- Vendor twin check (deployment repo discipline): the vendored copies of `db_info.py` and
  `card_query_nodes.py` carry these hunks byte-identically.

---

## Follow-up: the `is:` vocabulary, and a quoted type that matched nothing

A broad differential sweep against api.scryfall.com found the `is:` vocabulary partial in a way
nothing reported: `is:promo`, `is:reprint`, `is:firstprinting`, `is:spell` and `is:foil` all
returned ZERO against thousands on Scryfall, and `t:"artifact creature"` returned zero against 360.
No error, no warning — an empty result indistinguishable from a corpus with no such card. These
land here because they are parser-layer; the engine twins are on #907 and #927.

### Two of the five are derivable

`is:spell` — every card with a castable primary type on some face. Scryfall's is FACE-level and a
type union is not, so they differ on merged type lines: **+48 / 31,760, zero misses** on the
imported corpus (2026-08-16), the over-catch being single-faced Artifact Lands plus Unfinity's
Attractions. `-t:land` was the other candidate and is worse both ways (173 over, 87 under — it
drops the modal DFCs whose front face is a spell).

`is:firstprinting` — exactly `-is:reprint`, not approximately. `is:firstprinting is:reprint` and
`-is:firstprinting -is:reprint` are BOTH empty on Scryfall, so the two partition the printing
space; `e:khm` is 425 prints, 26 reprints, 399 first printings; `!"Lightning Bolt"` is 64, 61 and 3.
Ties all count (`!"Forest"` answers with both lea/294 and lea/295), which falls out of the
complement with no rule of its own. Plus `is:firstprint`, and Scryfall's second names for two land
cycles already carried: `is:karoo` (= bounceland, 17) and `is:canland` (= canopyland, 6).

### Three are booleans the bulk data already ships

`BOOLEAN_IS_TAGS` deferred `promo`/`reprint`/`foil` once over cardinality — "memory check first".
The check is now run rather than assumed: the Cloudflare port's builder over the same corpus,
twice, archives totalling **363.02 MiB without them and 364.17 MiB with — +1.16 MiB, +0.32%** for
three of the densest tags in the vocabulary, because a value carried by that share of the corpus
stores as a bitmap plane rather than a posting list. `foil` reads the deprecated top-level boolean,
which says the same thing as `finishes` containing `"foil"` and keeps every entry on the one shape
`_sync_boolean_is_tags` can express in SQL.

The dict moves to `db_info` because the parser reads it too. It is half of `SUPPORTED_IS_VALUES` —
the complete list of `is:` values this parser can answer.

### Everything outside that list now says so

An `is:` value with no data behind it gets a **warning on the parsed Query** instead of a silent
zero. Scryfall handles an unknown value by ignoring the term and warning (`is:notarealtag e:khm`
returns the whole set there); this parser keeps the term, so the answer is a no-match — and the
warning is what tells a caller which of the two happened. Adopting ignore-and-continue wholesale is
a separate decision that touches every operator, so it is not taken here.

A sweep of the whole documented `is:` vocabulary (129 values, both sides) puts ~55 more in that
"silent zero" bucket — mostly `promo_types` membership (`prerelease`, `buyabox`, `datestamped`,
`universesbeyond`, `boosterfun`, `showcase`, …) and further compat booleans (`nonfoil` 33,537,
`booster` 28,307, `hires` 32,970, `full`/`fullart` 3,228, `textless`, `variation`, `spotlight`).
All of those are now warned rather than silent, and each is a small follow-up in the same two
places.

### And the quoted type

`t:"artifact creature"` arrives as ONE title-cased token, "Artifact Creature" — neither a type nor
a subtype, so both containment branches answer false for every card. Scryfall matches the quoted
string against the whole type line as a case-insensitive substring; measured, not guessed: order
matters (`t:"creature artifact"` is empty), neither side is word-anchored (`t:"tifact creat"`
returns the same 360), whitespace runs collapse, the em dash is ordinary text (`t:"creature —
human"` matches, the hyphen spelling does not), and subtype pairs behave identically
(`t:"human wizard"` 6, reversed 0). Single-word values keep the membership path and its indexes;
the set-comparison operators keep theirs, since a substring is not a set. The engine twin is on
#907, where `TextField::TypeLine` lives.

### Validation

- `api/parsing/tests/test_rewrite.py` — the new equivalences, the vocabulary/importer agreement
  check, warned and unwarned `is:` values, warnings under negation and inside or-groups, and
  `t:` explicitly NOT participating.
- `api/parsing/tests/test_sql_gen.py` — the raw-tag lowering case moved to two tags that are still
  raw (`is:reprint`, `is:foil`), since `is:spell` is now a rewrite.
- Full parser suite: **1,791 passed, 19 xfailed**. `ruff check` / `ruff format` clean.
- End to end on the Cloudflare port against a policy-clean set (`e:dsk`, `unique=prints`):
  `is:spell` 381 = 381, `is:reprint` 23 = 23, `is:promo` 7 = 7, `is:foil` 417 = 417,
  `is:firstprinting` 394 vs 396 and `t:"legendary creature"` 34 vs 35 (the two printings that
  corpus excludes).

---

## Follow-up 2: the rest of the vocabulary, and one WRONG answer

The sweep above measured 132 `is:` values on both sides. This closes the two families whose data
the bulk file already carries, and fixes the single mapping that was returning a wrong answer
rather than a missing one.

### `is:new` was `frame:2015`, and that is not what Scryfall means

The comment in this file asserted `is:new` was "the 2015 frame *only* — deliberately narrower than
`frame:new`". Measured 2026-08-16 that is false in both directions:
`is:new -(frame:2003 or frame:2015 or frame:future)` is empty, its converse is empty, and
`frame:new -is:new` is empty too. **`is:new` IS `frame:new`, exactly.** The old mapping
under-matched by **9,201 cards**, every one of them a 2003 frame — Innistrad, Theros, Return to
Ravnica, the whole modern-frame era before 2015. This is the only wrong-answer mapping the sweep
found; the other five skews it flagged all turned out to be deliberate or corpus-policy (below).

### The booleans, and a second table for the arrays

`booster`, `digital`, `fullart`, `hires`, `nonfoil`, `spotlight`, `textless`, `variation` join
BOOLEAN_IS_TAGS. Seventeen more are membership in a bulk ARRAY rather than a boolean, so
`ARRAY_IS_TAGS` maps `tag -> (array key, member)` and `_sync_is_tags` — the old
`_sync_boolean_is_tags`, generalised — rebuilds both halves in one statement with jsonb's `?`
operator. Both managed sets are removed together before either is re-added, so a tag moving
between the tables converges in one import instead of surviving as a stale key.

Every mapping was established by READING the cards Scryfall returns rather than guessing the
spelling: `is:X` fetched, the `promo_types` arrays of the results intersected. That is what turns
`is:judge` into `judgegift` and what separates `is:stamped` from the broader `promopack` its
results also all carry.

### Twenty-one that were already expressible

`is:showcase` = `frame:showcase` (2,213 = 2,213) · `is:extendedart` = `frame:extendedart`
(3,629 = 3,629) · `is:tdfc` = `layout:transform` (both differences empty) ·
`is:bikeland`/`cycleland`/`bicycleland` = `otag:cycle-bicycle-land` (10) · `is:surveilland` =
`otag:cycle-mkm-surveil-land` (10) · `is:pathway` = `t:land name:pathway` (10 = 10) ·
`is:tricycleland` = the TRIOMES, name for name · the six unimported layouts
(`artseries`/`augmentation`/`host`/`planar`/`reversible`/`token`) · `is:full` and
`is:promostamped` as aliases rather than second stored copies.

### Cost, measured across four builds of the same corpus

| stored `is:` vocabulary | archives | vs 3 entries |
|---|---|---|
| 3 (reserved, gamechanger, oversized) | 363.02 MiB | — |
| 6 (+ promo, reprint, foil) | 364.17 MiB | +1.16 / +0.32% |
| 27 (full, minus booster/hires/nonfoil) | 370.79 MiB | +7.77 / +2.14% |
| 30 (full) | 372.68 MiB | +9.66 / +2.66% |

**The intuition the original deferral was built on is backwards.** `booster`, `hires` and
`nonfoil` are the three densest tags — 28k, 33k and 34k cards, the last of them 99.8% of the
corpus — and together they cost 1.89 MiB, a fifth of the growth. The 24 mid-density tags cost
6.61 MiB. Past the storage crossover a value is a bitmap plane (one bit per row) rather than a
posting list (four bytes per row), so the expensive tag is the one carried by a few thousand
printings, not by all of them. Nothing in either table is disproportionate on those terms.

### The five other skews: all deliberate or corpus-policy, none a bug

Diagnosed against the live API rather than assumed:

| skew | verdict |
|---|---|
| `is:dfc` 2,895 vs 507 | `is:dfc -layout:transform -modal_dfc -meld` is 2,394 on Scryfall, and filtering further by token / art_series / reversible_card empties it. Scryfall counts non-gameplay objects this import never carries; the layout union is right. |
| `is:oversized` 546 vs 10 | 204 are memorabilia, 331 are planar/scheme/vanguard, leaving 11 — of which this corpus has 10. Entirely the import filters. |
| `is:gainland` 15 vs 43 | Scryfall's own `otag:gainland` is 43. We match the community tag; their `is:` is a narrower curated list. Recorded deviation, confirmed. |
| `is:frenchvanilla` 1,095 vs 1,319 | Same shape: `otag:french-vanilla` is 1,352 there. |
| `is:storageland` 12 vs 15 | Same shape, already recorded as "15 vs 12". |

### The 43 values still warning, and why

- **Set-shaped** (`intro`, `masterpiece`, `alchemy`, `funny`, `scryfallpreview`): `is:masterpiece`
  is `st:masterpiece` exactly, but there is no set-type predicate. `set_type` IS stored.
- **Presence predicates** (`watermark`, `indicator`, `illustration`, `flavorname`): all four are
  Scryfall's `has:` family, not `is:`. Three of the four columns are stored; `flavor_name` is not
  ingested at all.
- **Derivable, but by a mechanism that does not exist yet**: `is:unique` (exactly one printing —
  a finalize-stage window over the printing count, neither a boolean nor an array),
  `is:default`/`is:atypical` (disjoint, union = every card — this is `default_cards` membership,
  which the multilingual import already carries as `is_canonical`), `is:localizedname` (the
  printed-name annex on #927), `is:hybrid`/`is:phyrexian` (mana-symbol classes over a column that
  is stored), `is:oathbreaker`/`is:brawler`/`is:duelcommander` (eligibility predicates in the
  shape `is:commander` already uses — each needs its own count validation).
- **Genuinely absent**: `is:etb` (not a community oracle tag — `otag:etb` and
  `otag:enters-the-battlefield` are both empty on Scryfall — so it would need an oracle-text
  heuristic), `is:meldpart`/`is:meldresult` (the face merge makes a meld card one row, so the
  part/result split is gone), `is:partner` (`is:partner -kw:partner` is 104, so the keyword is not
  it).

---

## Follow-up 3: two missing operators, not missing data

Of the 28 `is:` values still answering zero after the last pass, most were not missing DATA — they
were missing an **operator**. Two of them close 18.

### `st:` — the set type

Scryfall documents it, the bulk ships `set_type` on every card, and the engine has stored it since
the compat blob existed. Five `is:` values ARE set types: `is:masterpiece` = `st:masterpiece`
**exactly** (both set differences empty, 2026-08-16) and `is:alchemy` = `st:alchemy`. So this adds
the operator rather than five more stored tags. `is:funny` maps too, at 151/190 rather than exactly
— unobservable here, since the funny sets are not imported, and the mapping is what turns an
unexplained zero into an honest one.

The engine leaf is `LangMatch`'s shape verbatim (compat blob, `coll_vocab`, id resolved in `bind`,
one integer equality in `tri`) and lives on #927. Equality only, as on `lang:`; the hyphenated
spelling folds to Scryfall's underscored one.

**Flagged, not assumed:** `magic.cards.card_set_type` does not exist yet. This is `lang:`'s exact
situation on this branch — parser and engine here, column with the store PR.

### `has:` — the presence family

`has:` shares `is:`'s column, so an unmapped value was the same silent zero. The vocabulary was
read off the **live API**, not the syntax docs, which list two of it. It has two shapes and both
were already expressible: the boolean half (`has:foil`, `has:booster`, `has:highres`, …) asks
exactly what `is:` asks and answers off the same stored tag; the presence half is a non-empty test
on a text column, which `<field>:/./` already says.

Not mapped, and warning for a stated reason: `has:illustration`, `has:stamp`, `has:multiverse`,
`has:tcgplayer`, `has:cardmarket`, `has:image`, `has:printedname`, `has:indicator` are presence
tests on ids, interned compat scalars or the annex — no regex path — and `has:attraction_lights`
/ `has:partner` have no stored column at all. Each needs a presence predicate in the engine.

### Eligibility

`is:oathbreaker` / `is:brawler` / `is:duelcommander` follow `is:commander`'s shape, each validated
against its own live list rather than rewritten to the format filter: they are strict SUBSETS of
`f:oathbreaker` / `f:brawl` / `f:duel`. Every card Scryfall names is matched (the "is: minus shape"
difference is ZERO in all three); the shapes over-catch by 15 / 27 / 121 on 287 / 2,318 / 3,323,
and each format's banlist does not close it, so the residual is recorded.

### Verified

`st:masterpiece` 869 = 869 globally; on `e:dsk` with `unique=prints`: `st:expansion` 419 = 419,
`has:watermark` 3 = 3, `has:artist` 419 = 419, `has:flavor` 177 = 177, `is:oathbreaker` 4 = 4,
`is:duelcommander` 42 = 42. Parser suite 2,005 passed / 19 xfailed, ruff clean.

### What is left, and why

**A premise worth correcting:** `is:default` / `is:atypical` are NOT `default_cards` membership.
Measured — they partition the PRINTING space (intersection 0, union every card) and `is:default` is
**81,499 of ~105,000 printings**, so it cannot be one-printing-per-card. It is "not a variant
printing", and the canonical flag already imported answers a different question.

The rest, with what each needs: `is:unique` (one printing — a finalize-stage window, neither a
boolean nor an array); `is:localizedname` (the printed-name annex, so #927's space);
`is:hybrid` / `is:phyrexian` (mana-symbol classes — `mana_cost_jsonb` stores each symbol as a key,
so an OR over the ~25 hybrid symbols would express it, but `m:/…/` is not a path the engine has and
each spelling needs its own validation); `has:` presence over ids as above. Genuinely absent and
staying warned: `is:etb` (not a community oracle tag — `otag:etb` and `otag:enters-the-battlefield`
are both empty on Scryfall), `is:meldpart`/`is:meldresult` (the face merge makes a meld card one
row), `is:partner` (`is:partner -kw:partner` is 104), `is:intro` and `is:scryfallpreview` (no
shared facet on their result sets at all).

## `fo:` and `fulloracle:`

Scryfall's FULL-oracle operator, which this parser had no spelling for at all — both 400ed as
unknown operators. Measured against api.scryfall.com 2026-08-16:

| query | `o:` | `fo:` |
|---|---|---|
| `"damage dealt by this creature also causes"` | 0 | 71 |
| `"you may pay an additional"` | 0 | 268 |
| `"level up only as a sorcery"` | 0 | 25 |
| `draw e:khm` | 39 | 57 |
| `/\(/` | 0 corpus-wide | 148 on `e:khm` |

**They share `oracle_text`'s column, deliberately.** The stored `oracle_text` is already the full
text — reminder text included — so the SQL path answers `fo:` from it directly and needs no second
column, no migration and no index. What tells the two apart is `original_attribute`, which the wire
tree already carries: it matters only to the card engine, whose searchable oracle column has the
reminder text stripped out of it the way Scryfall's `o:` does (#907).

The consequence worth stating plainly: **on the SQL path `o:` and `fo:` are the same query**,
because `o:` there has never stripped anything. That is the pre-existing divergence, not a new one
— this only gives the operator that WANTS the full text a name for it.

`flavor:`/`ft:` were checked rather than assumed and are unaffected: `ft:/\(/` returns 47 rows on
api.scryfall.com, so flavor text is not stripped on either side.

## The mana-symbol classes, and the two values the engine answers

Four more `is:` values that returned a silent zero. `is:hybrid` and `is:phyrexian` are closed here
outright; `is:localizedname` and `is:unique` are closed by the engine and need only the vocabulary
entry that stops this parser reporting them unsupported.

### `is:hybrid` / `is:phyrexian` — symbol SET membership

The sets come from Scryfall's own `/symbology` filtered to `represents_mana`, not from a shape
guess. A symbol is **hybrid** when it has two or more non-Phyrexian components, and **Phyrexian**
when one of its components is `P`:

| class | symbols |
|---|---|
| hybrid (30) | the ten two-colour `{W/U}…`, the ten two-colour Phyrexian `{G/W/P}…`, the five colourless `{C/W}…`, the five monocolour `{2/W}…` |
| Phyrexian (16) | the ten two-colour Phyrexian, the five mono `{W/P}…`, and `{C/P}` |

They overlap on the ten two-colour Phyrexian symbols and part company on `{B/P}` and `{C/P}`, which
are Phyrexian and **not** hybrid — one colour each.

Both rules were verified against api.scryfall.com **card for card**: all 603 `is:hybrid` and all 73
`is:phyrexian` fetched and diffed against the 2026-08-16 bulk. **Zero** cards Scryfall names are
missed by either rule. Every extra the rules would add comes from a set an import may not carry
(Unknown Event, Mystery Booster playtest, Heroes of the Realm, 2017 Heroes of the Realm).

**`m:` leaves, not a regex over the printed cost.** The cost is stored as counted symbols, so each
leaf is an integer compare against the mana vocabulary; a regex over the cost string was measured
wrong in both directions — 5 under, 35 over.

**The `o:` half of `is:phyrexian` is not decoration.** Scryfall's rule is the symbol ANYWHERE on the
card, not only in the cost: **36 of its 73 cards carry no Phyrexian symbol in any cost at all** —
Spellskite, the Souleaters, Hex Parasite, and every `{2}{B/P}: transform` back face. Dropping the
`o:` half leaves half the answer behind.

`is:hybrid` is cost-only by the same measurement, in the other direction: **216 cards carry a hybrid
symbol in their rules text and Scryfall calls none of them hybrid.** The asymmetry is Scryfall's,
and matching it is why the two expansions are shaped differently.

### `ENGINE_IS_VALUES`

A third source for `SUPPORTED_IS_VALUES`, beside `BOOLEAN_IS_TAGS` and `ARRAY_IS_TAGS`. These are
values no rewrite can express and no importer tag holds — the engine answers each from a field it
already stores — so without the entry they would work and still warn that they do not, which is the
exact defect `SUPPORTED_IS_VALUES` exists to remove, in reverse.

* **`localizedname`** — "this printing carries a printed name". Presence, not difference and not
  "non-English": 182 of the printings it matches on api.scryfall.com are ENGLISH (om1/66 prints
  "Rhilex the Accursed" over Agent Venom), 4,468 foreign printings print a name byte-identical to
  the English one and still count, and it reads per FACE. `has:printedname` is its other spelling
  and both answer 31,294 there, so the `has:` table gains the alias and loses its "no predicate
  exists" note.
* **`unique`** — "this card has been printed in exactly one SET", which Scryfall's syntax page says
  in as many words. It is **not** `prints=1`: 2,847 of its own 16,318 matches have more than one
  printing (`!"Forest"` is the shape — two lea printings, one set).

The engine side of both is on #927. A `test_engine_answered_is_values_are_supported_and_stored_nowhere`
pins the other direction too: neither value may become an importer tag or a rewrite as well, or the
engine leaf would silently win over a stored tag nobody could observe.

### Measured on the Cloudflare port

Against api.scryfall.com over a store rebuilt from the 2026-08-16 all_cards bulk, exactly equal on
every policy-clean set probed: `is:hybrid e:mh3` 10 = 10, `is:phyrexian e:one` 13 = 13,
`is:unique` e:dsk 185/185, e:khm 142/142, e:one 156/156, e:mh3 175/175, `is:localizedname e:dsk`
276 cards / 1,917 printings on both sides. Corpus-wide, `is:hybrid` is 596 against 603 and
`is:phyrexian` 72 against 73 — the 8 misses are cards this import does not carry at all (Unglued,
Unstable, Unknown Event, and the unreleased `sos` prepare cards).

---

## Also on this branch: the colour NAMES, and the quotes every keyboard produces

Two more vocabularies the parser did not have, both of them things a user types by accident rather
than on purpose. Same class of change as the operators above — a table, not a mechanism — so they
ride here rather than opening a fourth branch.

### `c:azorius` was a parse error, and answers 6 cards on Scryfall

The parser knew six colour words — white/blue/black/red/green/colorless — and bare letter strings.
The ten guilds, the ten shards and wedges, the four-colour names, `rainbow`, `all`, `brown` and the
British spellings were all 400s. `COLOR_ALIAS_TO_CODES` adds the 44 names as the LETTER SET each
one spells, so `c:azorius` and `c:wu` serialize to the identical rhs and cannot drift apart, and
`get_colors_comparison_object` expands the name before its single letter path rather than beside it.

Measured one request each against api.scryfall.com (2026-08-16), then every accepted name checked
against its letter spelling **over the whole corpus** — Kaldheim holds exactly one card of three
colours or more and would have agreed with almost any mapping:

| name | letters | cards |
|---|---|---|
| `c:bant` | `c:gwu` | 153 |
| `c:esper` | `c:wub` | 146 |
| `c:yore-tiller` | `c:wubr` | 62 |
| `c:witch-maw` | `c:gwub` | 63 |
| `c:rainbow` | `c:wubrg` | 60 |
| `c:brown` | `c:c` | 4,300 |

…and so on for all 24 pairs. `id:`/`identity:`/`ci:` take the same vocabulary, measured separately
rather than assumed (`id:bant e:khm` = `id:gwu e:khm` = 176).

**It is a boundary, not a superset** — the half worth reviewing. `yore`, `glint`, `dune`, `ink` and
`witch` on their own are REJECTED by Scryfall, so only the hyphenated four-colour forms and the five
one-word synonyms are in its table; `five`, `mono`, `guild`, `shard`, `wedge`, `nephilim` and
`chromatic` are rejected too. Adding a name Scryfall does not have would answer a wider result than
Scryfall, silently, so the rejections have their own test.

`all` spells `wubrgc` where `rainbow` spells `wubrg`, and the difference is measured rather than
cosmetic: the `c` drops out of a card_colors comparison (`c:all` = `c:wubrg` = 60) and does not drop
out of produced_mana (`produces:all` matches nothing; `produces:rainbow` = `produces:wubrg` = 13).
One table serves all three columns because `get_colors_comparison_object` already draws that line.

The hyphen needed lexer work: `-` is not a word-continuation character, so `c:yore-tiller` arrives
as WORD MINUS WORD. `parse_color_value` glues the three back together only when the result is a
name, using the same adjacency rule `parse_date_value` uses for YYYY-MM-DD, so an ordinary `c:w-1`
still ends its value at the `w`.

**Deliberately NOT included: `m` / `gold` / `multicolor(ed)`.** Those are a colour COUNT, not a set —
`c:m` is `c>=2` (44 in Kaldheim) where `c:2` is 43, and `c<m` is `c<2` while `c:m`, `c=m`, `c>=m`
and `c>m` are all `c>=2`. They need the numeric colour-count value node, which is #879's and not
this branch's; landing them here would mean carrying that prerequisite along and conflicting with it.

### The four typographic quotes

Every word processor and phone keyboard turns a typed apostrophe into U+2019 and typed double quotes
into U+201C/U+201D. This parser read them as ordinary letters, so `name:"Gaea’s Blessing"` searched
for a name containing a curly apostrophe: no rows, no error, no clue. `fold_typographic_quotes` now
folds the four Scryfall folds — in `parse_query` and `parse_search_query` before the lexer, and in
`balance_partial_query` so the balancer and the lexer agree about what a quote is.

Measured by putting each candidate around a phrase and asking whether the phrase searched as ONE
term. U+2018/U+2019 fold to `'` and U+201C/U+201D to `"`; every other quotation-shaped character
stays literal and matches nothing — guillemets, the low-9 pair, the primes, the fullwidth forms, the
CJK brackets, the ornate pairs, backtick, acute, U+02BC. It is a CHARACTER substitution over the
whole query rather than a rule about quoted regions, and **both directions had to be measured**:
folding all four to `"` fits `name:"Gaea’s Blessing"` finding the card and fails
`name:‘Gaea"s Blessing’` finding nothing.

### Verification

`api/parsing/tests/test_color_names.py` (44 name/letter pairs, the case-insensitivity cases, the
twelve rejections) and `api/parsing/tests/test_typographic_quotes.py` (the fold, the 22 characters
that do not fold, the balancer) are new. Both parsers take both changes — pyparsing's colour word is
built from the same table and folds in the same position — so `test_parser_parity` keeps them from
diverging. `ruff check` and `ruff format` clean; the parsing suite is 2,036 passed / 19 xfailed.

On the Cloudflare port carrying the same hunks, against api.scryfall.com: `c:azorius e:khm`,
`id:bant e:khm`, `c:yore-tiller e:khm`, `c:all`, `produces:all`, `c:brown`, `c:AZORIUS`,
`c:"azorius"`, `c>=azorius`, `c=azorius`, `c<azorius`, `c<=azorius`, `c!=azorius`, `-c:azorius`,
`produces:azorius` and the curly-quote forms of `name:` and `o:` all became exact matches, and six
new cases in the live-parity harness compare byte-for-byte against api.scryfall.com.

---

## Follow-up (2026-08-16): a ligature is not an accent

`fold_accents()` is NFKD plus a combining-mark filter, and that is a *decomposition*: it can only
ever separate a base letter from its marks. `æ` is not `a` with a mark on it — it is its own letter,
with no decomposition at all — so it, and `œ`, `ß`, `ø`, `ł`, `đ`, `þ`, `ð`, `ħ`, `ŋ`, `ŧ`, `ı` and
`ĸ`, passed through untouched. `name:ætherling` found nothing; so did `name:straße`, and `a:gawel`
for Bartłomiej Gaweł.

Scryfall expands all thirteen. Measured on api.scryfall.com, one probe per character against its
expanded spelling, equal totals on both sides of every pair:

| | | | | | | |
|---|---|---|---|---|---|---|
| æ/ae | 90 | œ/oe | 167 | ß/ss | 2,051 | |
| ø/o | 22,111 | ł/l | 18,748 | đ/d | 14,591 | þ/th 5,689 |
| ð/d | 14,591 | ħ/h | 14,176 | ŋ/ng | 4,834 | ŧ/t 22,261 |
| ı/i | 22,954 | ĸ/k | 6,616 | | | |

(`ĳ` folds too, at 22 — NFKD already reaches that one through its *compatibility* decomposition, so
it is not in the table.)

`_LIGATURE_FOLD` is applied AFTER the mark filter, so nothing about the existing behaviour moves:
every case in `test_fold_accents` passes unchanged, and twelve new ones pin the expansions.

Three characters are deliberately absent, each **measured to 404 on Scryfall** rather than assumed:
`×` and `÷`, which are symbols rather than letters and which `collate_name()` would delete in any
case; and `ſ`, which Scryfall does not fold and NFKD does. The remaining residuals — `ſ`, the
presentation ligatures `ﬁ`/`ﬂ`/`ﬀ`, `½`, `№`, `ǽ` — are all NFKD folding where Scryfall does not, on
characters no card in the corpus contains; they are documented on the constant.

`fold_accents()` precomputes `card_name_folded` at import as well as folding the search term, so
both sides of the comparison move together and no reindex is needed beyond the usual re-import.

On the Cloudflare port carrying the same hunk, against api.scryfall.com: `name:æther` 90 (was 0),
`name:œuvre`, `name:straße`, `a:łebski` 6 and `a:gawel` 10 all became exact matches.

---

## Follow-up (2026-08-16): `is:dfc` is not the double-faced union, and two more layout values were half a predicate

`is:dfc` was `layout:transform or layout:modal_dfc or layout:meld`, with the comment saying Scryfall
"additionally counts art_series / reversible_card / double_faced_token (~2394 art & token entries)
that aren't gameplay cards and aren't in our corpus, so the layout union is the correct set for our
data."

**Both halves of that are false now, and the mapping is wrong in BOTH directions.** Measured against
api.scryfall.com on 2026-08-16, on two independent axes — `unique=prints` with
`include_extras`+`include_variations`, and the plain default `unique=cards` — `is:dfc` is exactly

```
layout:transform or layout:modal_dfc or layout:art_series or layout:double_faced_token or
layout:reversible_card
```

| | prints (+extras) | cards (default) |
|---|---|---|
| `is:dfc` | 4,244 | 2,895 |
| `is:dfc -(the five layouts)` | **0** | **0** |
| `(the five layouts) -is:dfc` | **0** | **0** |
| `is:meld is:dfc` | **0** | **0** |
| `is:meld -is:dfc` | 72 (all of them) | 21 (all of them) |

So it **excludes meld outright** — which the old union included — and reaches 2,851 printings the old
union did not: art_series 2,650, double_faced_token 120, reversible_card 81.

**"Not in our corpus" is what makes this a live defect rather than a definition footnote**, and it
stopped being true when `preprocess_card` stopped DROPPING the extras classes and started TAGGING
them (`_EXTRA_LAYOUTS` / `is:extra`, this branch's sibling work). Counting `layout` over the
2026-08-16 all_cards bulk the import now reads gives **art_series 2,650 and double_faced_token 120 —
in exact agreement with Scryfall.** Every `is:dfc` query answers 2,770 printings short today, and
answers 72 it should not.

### The rest of the layout family, probed the same way

Two more entries were half a predicate each:

* **`is:host` and `is:augmentation` are the SAME predicate**, and it is Unstable's two halves
  together: `is:host -is:augmentation` and its converse are both empty, and each equals
  `layout:host or layout:augment` (46 = 29 + 17). They were one layout each, so each was missing the
  other's — `is:host` short by 17, `is:augmentation` short by 29. Written out per value rather than
  aliased to each other, because their equality is a measurement about Scryfall, not a spelling.
* **`is:token` reaches past `layout:token`** to the double-faced tokens (120) and to six Wilds of
  Eldraine Role tokens that ship as `layout:flip` (twoe/15–17, twoc/1–2, plst/TWOE-17). `t:token` is
  the clause that catches those six and it cannot over-catch — `t:token -is:token` is empty — so
  `layout:token or layout:double_faced_token or t:token` is exactly `is:token` on both axes, and
  stays exact as further odd-layout tokens are printed. (`is:token layout:emblem` is 0: an emblem is
  not a token.)

`is:mdfc`, `is:tdfc`, `is:transform`, `is:meld`, `is:artseries`, `is:reversible`, `is:split`,
`is:flip`, `is:leveler`, `is:adventure` and `is:planar` were probed identically and are exact in both
directions. They are unchanged.

This is **independent of the `include_extras` auto-enable table**, which keys on the TERM the caller
wrote rather than on what it expanded into — `is:dfc` fires there either way, before and after.

### Verification

`test_rewrite.py` pins the four new equivalences, including `is:host == is:augmentation` as two rows
against the same DSL. On the Cloudflare port carrying the same hunks, against its own store:

| | before | after | api.scryfall.com |
|---|---|---|---|
| `is:dfc` (prints) | 1,460 | **4,158** | 4,244 |
| `is:token` (prints) | 2,988 | **3,114** | **3,114** |
| `is:host` | 29 | **46** | **46** |
| `is:augmentation` | 17 | **46** | **46** |

`is:token`, `is:host` and `is:augmentation` are now exact. `is:dfc`'s residual 86 is not this
mapping: 81 are `layout:reversible_card` printings that the face merge is mislabelling (Scryfall puts
a `layout` key on the FACES of reversible cards, the only 81 cards in the whole bulk that carry one,
and the face overlay overwrites the printing's `reversible_card` with the face's own `normal`/
`adventure`/`token` — filed separately against the import), and 5 are transform/modal_dfc printings
newer than that port's store.

## Added: `is:vanilla` expanded to a tautology (90801b0)

`is:vanilla` is **363** on api.scryfall.com and the expansion answers **18,753** — which is exactly
`t:creature`, the other half of the expansion. The empty-oracle conjunct never narrowed anything.

It is not the engine dropping a term, and the diagnosis is the one shape of three that is **not** a
filter bug. `o=""` is present, is compiled, and is vacuous — on **both** sides. Measured
2026-08-17: `o=""` and `o:""` are each a **400** on api.scryfall.com ("All of your terms were
ignored"), and `t:creature o=""` answers **18,753** there, the same number this expansion produces.
The term asks for the empty string as a SUBSTRING, and every string contains it.

The expansion was written when `=` on a text column was an equality. It is not one any more, and
that was itself a measured parity fix (`o=flying` = `o:flying` = 4,574 on Scryfall, where a real
equality answers 99). This is the term left behind by it.

The empty-text test that does exist is the presence regex the `has:` family already uses, negated:
`t:creature -o:/./` answers **352** on api.scryfall.com and **352** on this corpus — the two agree
on the spelling, which is what makes it a replacement rather than a second guess. Scryfall emits an
`oracle_text` key on every single-faced creature (0 of 20,098 in the oracle bulk lack one), so the
negation is not exposed to SQL's three-valued `NOT` on the way through.

352 and not 363, and the 11 are one class: Scryfall's `is:vanilla` is **face-level**. All 12 rows
of its own `is:vanilla o:/./` are adventures and their kin, whose CREATURE face prints no rules text
while the other face does (`Beluna's Gatekeeper // Entry Denied`). The stored `oracle_text` is the
merged row — every face's text joined — so the regex sees the other half. Closing that needs a
face-scoped predicate, which is not something a rewrite can reach.

`has:vanilla` moves with it, since `has:` is a total alias of `is:` (this PR's previous commit).
`is:vanilla` was the only expansion in the map spelling an empty value; the rest of the `is:`
vocabulary was audited conjunct by conjunct and none of it shares the shape.


daveycodez:
### `commander:` and the British spellings

Three alias spellings Scryfall accepts that this parser answered with a bare NAME search — the
quiet kind of wrong. `commander:wu` did not filter by colour identity; it looked for cards *called*
"commander:wu", found none, and said so without complaining.

Measured one request each against api.scryfall.com, 2026-08-16, scoped to Kaldheim:

```
commander:wu = id:wu = identity:wu = ci:wu = 117
colour:wu    = colours:wu = color:wu = colors:wu = c:wu = 6
```

`commander:` is plain colour **identity** — not a variant of it, not `is:commander`. It inherits the
numeric and colour-COUNT forms for free, because the lowering keys off the resolved db column rather
than the spelling: `commander:m` = `commander>=2` = 74, and `colours:m` = `colour>=2` = 44. (Those
count forms arrive with #879; the alias works for letter values on this branch alone.)

### The vocabulary is a boundary, and it was measured as one

`cid`, `commanderidentity`, `colouridentity` and `colour_identity` all come back `Unknown keyword`
from Scryfall, so none of them joins the table — and a test pins that they stay out. An identity
alias this parser invents answers a **wider** result than Scryfall for exactly the filter people use
to check whether a card is legal in their deck, which is the direction that actually hurts.

`color_identity` and `coloridentity` are the mirror image — spellings this parser accepts and
Scryfall does not — and are deliberately left alone. Answering where Scryfall merely warns costs a
searcher nothing, while removing them would break queries that already work.

### Verification

`pytest api/parsing/tests` 2,056 passed / 19 xfailed. `ruff check` / `format --check` clean.

daveycodez:
Added `93352c7b`: the same 82 `promo_types` rows, meld roles and spelling aliases that #1073 adds to main, applied to this branch's copy of `BOOLEAN_IS_TAGS` (which lives in `db_info.py` here), plus `atypical`/`default` in `ENGINE_IS_VALUES` for the leaf on #912. Also carries #1073's fix for the `jsonb_build_object` 100-argument cap, which any table past 50 rows hits.

PR #927, opened 2026-08-16
Multilingual Store: Foreign Printings From all_cards

daveycodez:
> [!IMPORTANT]
> **Stacked.** This branch builds on `engine-fuzzy-name-scan`, whose one commit already rides
> #912's branch (patch-identical to b5bf12d there) — so the diff here includes the #894 → #913 →
> #912 stack until those land. Merge order: #894 → #913 → #912 → this. The `lang:` parser operator
> (`parser-lang-operator`, based on main) is independent and should merge before or alongside —
> the SQL lane's `card_lang` column ships here, the operator there.

## Motivation

The store indexes English names and printings only, because the import feeds `default_cards`.
"ego à deriva" — which Scryfall resolves to the Portuguese printing of Unmoored Ego — is
unanswerable, `lang:` doesn't exist, and `include_multilingual` is a no-op. all_cards is 540,484
printings (114,068 in default_cards; ja 62k, fr 59k, de 58k, es 53k, it 53k, zhs 41k, pt 40k,
zht 24k, ru 22k, ko 15k + tail). This PR makes foreign printings native rows — searchable by
`lang:`, resolvable by printed name at printing granularity — without the default (English) lane
paying anything for their existence.

## The load-bearing shape: an annex, not wider ranges

`CardData` gains `foreign` + `foreign_offsets`: non-canonical printings, CSR by the same cards in
the same prefer-desc build order. A separate range rather than a suffix inside `printings`,
because an in-range suffix would grow every printing-space bitmap plane ~5.7x and tax every
default query — the same disease as the memorabilia conjunct this repo already declined once —
and every `0..n_printings` loop in a 14k-line file would need auditing. With the annex, the
canonical space (every index, all six physical plans) is provably untouched: default queries
never read the new arrays. A **virtual pid** addresses both spaces (`vpid < printings.len()` is
canonical; anything else indexes the annex), with one resolution function (`printing_at`) so the
split cannot be interpreted two ways.

Canonical is **id-membership in Scryfall's own default_cards** — fetched alongside all_cards for
its ids, stamped per row at import (`is_canonical`), never re-derived from per-row facts:
default_cards names a few non-English printings canonical (cards never printed in English), and
re-deriving Scryfall's choice is a drift class of its own.

## Commits

1. **Engine: A Foreign-Printing Annex, Printed Text on the Printing, and the Artist as Printed** —
   schema + build. `Printing` gains Scryfall's `printed_name` / `printed_type_line` /
[...]

PR #929, opened 2026-08-17
Engine: Layout Belongs to the Printing, and a Reversible Printing Prints Its Own Faces

daveycodez:
> [!IMPORTANT]
> **Stacked on #927.** This branch is based on `multilingual-store` at `1b50f87`, whose last commit
> ("Import: A Reversible Card's Layout Came From Its Front Face") is the direct prerequisite: it
> fixes the layout VALUE the 81 `reversible_card` rows carry, and this fixes the SHAPE that keeps
> that value from being able to answer anything. Merge order: #894 → #913 → #912 → #927 → this.

One behaviour: **a reversible printing answers for itself.** Part 1 moves `layout` off the oracle
card so the field can differ between printings at all; Part 2 gives those printings their own faces,
their own top-level name, and their faces' layout — without which `layout:` is still wrong by 77
rows and `card_faces` is empty on all 81 of them. Split across two PRs a reviewer would merge
"layout is per-printing but still wrong for reversible printings" and then a follow-up repairing it,
so they are here together. Two later commits ride along under **Added** at the end: the released-date
tiebreak, and the exact-name fix that closes the first item on this PR's own "what this leaves" list.

---

# Part 1 — layout belongs to the printing

## The defect

`card_layout_id` lives on `OracleCard`. Layout is a property of a piece of cardboard, not of an
oracle card, and on this corpus the difference is not academic.

Every `reversible_card` is a **second printing of an ordinary card**. All 71 oracle ids behind the
81 reversible printings also carry normal printings — Temple Garden has 20 printings, of which
exactly one is reversible. So the group's single layout string was always decided by whichever
printing the commit pass reached first, and it was never the reversible one. `is:reversible` could
not answer 81, or 71, or anything: it answered whatever that accident produced.

#927's last commit fixed the value those 81 rows carry. It could not fix the answer, because one
card cannot hold two layouts.

## Scryfall answers per printing, and says so

Measured against api.scryfall.com, 2026-08-16 — in the one place card-level and printing-level
storage differ observably:

| query | `unique=cards` | `unique=prints` |
|---|---|---|
| `is:reversible` | 71 | **81** |
| `layout:reversible_card` | 71 | **81** |
| `!"Temple Garden"` | 1 | 20 |
| `!"Temple Garden" is:reversible` | 1 | **1** |

The last row is the whole argument. Temple Garden has 20 printings and one of them is reversible.
A card-level field can answer that query with 0 or with 20. Scryfall answers **1**.

## What moves

`card_layout_id` moves from `OracleCard` to `Printing`, beside `card_border_id` and
`card_watermark_id` — which is also where the query side already groups it
(`card_set_code | card_layout | card_border | card_watermark | collector_number` are one arm in
`build_binary`).

**The archive pays nothing for this half.** The obvious estimate — 4 bytes across every printing —
is what makes it look like a budget question. Measured instead: the `u32` lands in padding the
printing's 16-byte (`u128`) alignment had already left, and vacating its slot on the card only
widens padding there. Neither archived row changes size.

---

# Part 2 — a reversible printing prints its own faces, its own name and its faces' layout

## The defect

Moving the field was not enough. Two more things about those 81 printings still came off the
`OracleCard`, whose values the group's FIRST row supplies — and that row is one of the ordinary
printings:

- they carried **no `card_faces` at all**, where Scryfall gives them two;
- they answered to **"Temple Garden"** rather than Scryfall's **"Temple Garden // Temple Garden"**.

And `layout:normal` read **106,558** against api.scryfall.com's **106,635** — the whole 77-row gap
being those printings' *faces*, which carry a `layout` of their own.

## The measurement that says face text is not oracle-level

`OracleFace` says face text is oracle-level "because every printing of a card prints the same
faces". Measured field for field over every printing of every oracle group in the 2026-08-16
all_cards bulk, that holds for **38,555 of 38,626** groups and fails for exactly the **71** with a
reversible printing:

- **67** whose ordinary printings are single-faced (Temple Garden, 63 of its printings) while the
  reversible one has two faces the card itself never prints;
- **3** `adventure` cards whose reversible printing prints the SAME two face names with the front
  face's oracle text and power/toughness copied onto the back — so the divergence is not just "more
  faces", and **no rule derives one list from the other**. A "take the row with the most faces"
  heuristic corrupts exactly these three;
- **1** `token` (Mechtitan).

The same 71 groups, and only they, disagree on the top-level `name`. The adventure cases are also
why the joined name is **stored** rather than derived by joining the face names:
`Bloomvine Regent // Claim Territory` against a THREE-part
`Bloomvine Regent // Claim Territory // Bloomvine Regent`.

## Where the record lives, and why

`OracleCard` gains `divergent: Vec<DivergentPrinting>` — the group's OTHER `(name, faces,
face_layout)` triple. **On the card, not on the printing**, for two reasons:

1. **It is card-level data.** The pair of face lists is a property of the oracle group — one card,
   two things its printings print — and *which* printings print which is decided exactly by their
   own `card_layout_id`, so a discriminator on the row would be a second copy of a fact the row
   already carries.
2. **It is the cheap side**, measured on this tree rather than assumed. Both archived rows are
   16-aligned. `Archived<OracleCard>` goes **304 → 320** for one 8-byte `Vec` header — ~0.6 MB over
   38,626 cards — where ONE byte on `Printing` would round *that* row 304 → 320 too and cost
   **~8.6 MB** across ~540k printings.

The layout key is **exact and not a proxy**: 0 of the 71 divergent groups have a layout that appears
on both of their pairs (the reversible printings are `reversible_card`; their siblings are `normal`,
`adventure` or `token`, never both). The commit pass **re-checks it per group** and errors rather
than storing a record that would answer for the wrong printings — and errors again on a third
distinct pair, which this shape cannot hold.

A `Vec` of 0 or 1 rather than an `Option`: rkyv archives either as an 8-byte header, the build is a
`push`, and a corpus that ever grew a third pair is expressible rather than a panic.

## `layout:` is a multi-VALUE column

That is what Scryfall's own answers say it is. Re-measured against api.scryfall.com at this sitting:

| query | prints |
|---|---|
| `layout:normal` | **106,635** |
| printings whose OWN layout is `normal` | 106,558 |
| `is:reversible` | 81 |
| `is:reversible layout:reversible_card` | **81** |
| `is:reversible layout:normal` | **77** |
| `is:reversible -layout:normal` | **4** |

81 under one value and 77 under another, *the same printings*, is not something a scalar column can
produce. So `layout:` reads the union of `Printing::card_layout_id` and
`DivergentPrinting::face_layout_id`, existentially — and negation composes correctly through it,
which is what keeps `is:reversible -layout:normal` at Scryfall's 4 rather than 81.

## The card object

The 81 objects also stop carrying three keys Scryfall omits **on them alone**. A reversible printing
sends no top-level `oracle_id`, `cmc` or `type_line`, where a `transform` printing sends all three
(verified live on Delver of Secrets // Insectile Aberration). Its FACES carry the card's `oracle_id`
and `cmc` instead — the same values on both faces, 0 of 81 disagreeing. Verified live on
`sld/1079` and `ecl/351`.

Both writers move together — `card_object.rs::write_scryfall_card` and
`scryfall_compat/objects.py::to_scryfall_card` — since the engine answers a card when it can and
the SQL path answers it when the engine cannot, and a client must not be able to tell which one did.

## The importer

`_FACE_OBJECT_FIELDS` had no `layout` entry, so the face's own layout was dropped at import and
nothing downstream could have read one. It is now kept, first in the record, which is where
Scryfall puts it. #927's `card_layout` re-lift stays exactly as it is: the two are *different*
values on the same row (`reversible_card` at card level, `normal` on both faces), and both are
searched.

---

# Consumer enumeration

More than one structure decides whether a row survives, and the extra one is usually invisible to
the compiler. Every consumer of `layout`, of a card's face list, and of the top-level name, with
why it is or is not affected:

### Moved

| consumer | why |
|---|---|
| `filter::text_field_value` (`TextField::Layout`) | reads the printing, exactly as `Border` does |
| `filter::second_text_field_value` **(new)** | the second value; an **exhaustive `match`** over `TextField`, not a `matches!` with a hidden `_ => None` |
| `FilterExpr::TextExact` / `TextRegex` arms | existential over both values |
| **`ValueTotals::layout`** | see below — the one nothing forces |
| `FIELD_TABLE`'s `name` arm | the joined name on the 81, the card's name on everything else |
| `faces_to_pylist` | whichever face list this printing prints, plus the face `layout` key |
| `card_object.rs::write_scryfall_card` / `write_faces` | the reversible object shape |
| `scryfall_compat/objects.py::to_scryfall_card` / `_faces` | the Python twin of the same shape |
| `card_processing.py::_FACE_OBJECT_FIELDS` | keeps the face's `layout` instead of dropping it |
| `ARCHIVE_FORMAT_VERSION` | 2026081610 → 2026081611 |

**`ValueTotals::layout` is the structure a struct change cannot force.** It is read as an EXACT
total — the `TextExact { Layout, Eq }` arm of `exact_result_total` returns it and never runs the
filter — so a face layout missing from it is not a slower count but a **wrong** one. Leaving it
posting only the printing's own layout keeps `layout:normal` at 106,558 **while every filter
assertion above passes**. It is not a field on either archived row, so nothing breaks the build.
This is the second time this specific table has been the hidden consumer of a layout change.

### Checked and NOT affected

| consumer | why not |
|---|---|
| `estimator::has_printing_varying_leaf` | already classifies `Layout` printing-varying (Part 1); the second value is printing-level too, so the classification is unchanged and still correct |
| `filter::leaf_compares_printing_field` | same, and it is exhaustive, so it would have broken the build |
| `FIELD_TABLE`'s `layout` arm | emits the PRINTING's own top-level layout, which Scryfall agrees is `reversible_card` on these rows |
| bit planes | `compile_plane` has no `Layout` arm and there is no `PLANE_LAYOUT`; a layout leaf declines through the catch-all |
| `PairTotals` / `leaves_are_disjoint` | no `Layout` arm; `PAIR_MIN_PRINTINGS` prunes layout from the pair table entirely |
| every posting index in `CardIndexes` | layout has none, in either space — which is why this is a two-row change and not a ten-index one |
| `verify_cost_tier`, `cost.rs` | field-agnostic for `TextExact`; `cost.rs` names neither layout, name nor faces |
| the `is:` tag vocabulary | `card_is_tags` is already printing-space, and the `is:` layout predicates are rewritten to `layout:` in `rewrite.py` before the wire tree exists |
| `_merge_processed_faces` and the merged row's `card_name` | `preprocess_card` lifts the card's top-level name *before* the face overlay, so `magic.cards.card_name` already holds the joined name per printing. The divergence was created entirely by the engine's oracle grouping |
| `PrintingFace` / `printed_faces` (the ART half) | indexed positionally against the face list; a divergent printing's art faces are its own row's, already parallel |
| `raw_card_blob` readers (`copy_images_to_s3.py`, `prefer_weights.py`) | coalesce to `card_faces->0->'image_uris'`, untouched by a face record gaining a `layout` key |

### Affected and deliberately deferred

- **The SQL fallback's `layout:`.** `card_query_nodes.py` compiles `layout:` to
  `card.card_layout = %(p)s`, a scalar equality on a per-printing column. The column move in Part 1
  needs nothing there (`magic.cards` was always per-printing), but the multi-value half of Part 2
  does: the SQL path will keep answering `layout:normal` 106,558 where the engine now answers
  106,635. Making it match means `(card.card_layout = %(p)s OR card.card_faces->0->>'layout' =
  %(p)s)`, which drops `idx_cards_layout` on the second disjunct and needs its own measurement of
  `layout:adventure` / `layout:token` and of every `is:` rewrite that rides the same column. That is
  a separate change with its own test surface; flagging it rather than smuggling it in.
- **The `card_faces` backfill in `2026-08-10-01-engine-card-objects.sql`** strips only `object` and
  `image_uris`, so it *already* preserved a face-level `layout`. Adding the key to
  `_FACE_OBJECT_FIELDS` makes a re-import agree with a backfill instead of diverging from it. No
  migration needed.

---

# What this leaves

Stated rather than omitted:

- ~~**`!"Temple Garden // Temple Garden"` is 1 on Scryfall and 0 here.**~~ **Closed by the third
  commit below** — and it did not need the name indexes moved to printing space after all.
- **Face `artist_id` is missing on every multi-face card**, not only reversibles — `PrintingFace`
  stores the artist's name and vocab id but not the uuid. Pre-existing and independent of this
  change; not fixed here.
- **Face key ORDER.** The re-added keys go `object, oracle_id, cmc, …` where Scryfall puts `cmc`
  further along (after `mana_cost`). JSON object order carries no meaning, and the writer's existing
  rule is "re-added keys first" — but it is a difference, and the empty-value elision on faces
  already produces others (Scryfall sends `mana_cost: ""` on a reversible land face; both writers
  drop it). Left as-is so the two writers stay identical to each other.

---

# Tests

**Part 1.** `a_cards_printings_each_answer_with_their_own_layout` — Temple Garden reduced to three
rows, each answering for itself, with the exact-total table checked against the brute-force scan in
every space. `the_two_printing_dependence_tables_agree_on_every_text_field` pins the exhaustive
`match` and the silent `matches!` together over every `TextField`. And the **fuzz store now assigns
layout by (card index + printing index)** rather than by card, so the printings of one card
DISAGREE — a corpus where they always agree is one in which this entire class of bug is invisible.

**Part 2.** `a_reversible_printing_prints_its_own_faces_name_and_face_layout` — the same three rows,
now with a divergent record: which pair each printing prints, the joined name, `layout:` matching
under BOTH values on the reversible row, and the exact-total table answering 3 for `normal` where
posting only the printing's own layout would answer 2 while every other assertion still passed.

**Python.** `test_a_reversible_printings_faces_keep_their_own_layout` (the card's layout and the
faces' layout are two different values on one row, and the joined name survives the merge),
`test_an_ordinary_faces_record_carries_no_layout` (absence round-trips), and
`test_a_reversible_printing_keeps_nothing_of_the_card_at_top_level` (the object shape).

`ARCHIVE_FORMAT_VERSION` 2026081609 → **2026081611**: two struct-layout moves and two stored tables,
and an archive built by older code has them wrong. The `OracleCard` row size moves so the header
catches a stale archive — but `ValueTotals` does not, which is why the constant moves for the second
commit too.

---

## Added: a shared release date groups by set before it counts collector numbers

`order=released` broke a date tie by collector number alone, so the printings of every set that
shipped on a given day came back interleaved: a collector-number-only second key answers `tdc:1,
tdm:5, ptdm:5p, ptdm:5s, tdc:6 …` where api.scryfall.com answers all of tdm, then all of tdc, then
all of ptdm.

Set contiguity is measured, not assumed — over **102 sampled multi-set date groups** (five type
queries, both directions, `unique=prints`), the sets were contiguous in **102 of 102**, and on every
group tried `dir=desc` was the exact reversal of `dir=asc`, sets included.

So the second key becomes the pair (set, collector number), packed into the 32-bit lane it already
had as `(set_rank << 16) | collector_rank`. Complementing the packed word for `dir=desc` complements
each half in place, so the lexicographic pair reverses as one. **No stored bytes change** —
`order=released` has no prebuilt permutation and no walkable value index, so this is entirely a
query-time key.

### Disagree with this first: it is a deliberate approximation

The set order **inside** a date is Scryfall-internal and is not any field it publishes. `tdm`
precedes its commander deck `tdc`, but `mkc` precedes its expansion `mkm` — both verified directly —
so no parent/child, `set_type` or promo rule holds both. Nothing fit better than the set **code**
across `id`, `tcgplayer_id`, `card_count`, `printed_size`, `set_type`, `parent_set_code`,
`block_code`, the `/sets` listing order and the set name: the code fits **36 of 48** sampled adjacent
set pairs, the next best 30.

The cost is real and is not hidden. Replayed over 26 released-ordered pages (3,152 rows; every cached
listing at or above the eight-distinct-date floor, 24 carrying a multi-set date tie):

| second key | positional agreement |
|---|---|
| collector number alone | **86.96%** (2,741 / 3,152) |
| set then collector | **98.89%** (3,117 / 3,152) |

16 pages improve and **6 regress** — those whose only multi-set group is one where alphabetical
inverts Scryfall's block order (`tdm`/`tdc`, `sos`/`psos`, `4ed`/`4bb`). Their rows are still grouped
by set; two adjacent blocks swap. That residual **1.11%** is the whole of what Scryfall's private set
order costs this column, and closing it needs an order Scryfall does not publish.

### Two implementation notes

`assign_set_ranks` asserts the rank fits 16 bits (1,100-odd set codes against 65,535), so a corpus
that outgrew the lane is a **loud build failure** rather than a rank silently overwriting the number
beside it. The field stays `u32` — this is a contract on the value, not the layout.

The collector half needed no change and is confirmed rather than assumed:
`(collector_number_int, collector_number)`, where the int is *every digit concatenated*
(`extract_collector_number_int`), is the only collator with **zero violations** across khm, ptdm,
plst, sld and unf — 0 of 2,157 tied-date adjacent pairs, against 51 for a leading-digit int, 24 for
the first digit run, 51 for a plain string. `A-40` → 40, `M3C-368` → 3368, `★` → absent.

Also corrected in passing: `page_cmp`'s key list said "0 for every column but `set`", now stale, and
named key 3 as the edhrec rank where `sort_key_bits` puts `card.name_rank` there.

### Test

Written so the two candidate rules **disagree** rather than differ in confidence: the fixture's
`zza:500` sits above `zzb:1` ascending, which no collector-number-only key can produce at any
direction. Reverting the lane makes it fail with exactly `["zzb:1", "zza:9", "zzb:10", "zza:500"]`,
which is checked in as the sequence the old key gives. Ranks come from `assign_set_ranks` and
`assign_collector_ranks` rather than being written by hand, so the packing is exercised against the
values a build actually produces, and `dir=desc` is asserted to be the exact reversal.

The **rule** is engine-generic; only the cross-partition byte encoding is a downstream concern, and
upstream's SQL path has the same question to answer.

---

## Added: a reversible printing answers for the name it actually prints

`!"Temple Garden // Temple Garden"` returned nothing where api.scryfall.com returns the one
printing that prints it — the first item on the "what this leaves" list above. Same for
`!"Mechtitan // Mechtitan"` and `!"Adrix and Nev, Twincasters // Adrix and Nev, Twincasters"`: 81
printings over 71 cards, each printing a joined name its card does not have.

That item said closing this meant moving the name indexes into printing space. It does not. The
name indexes stay exactly where they are; the cost is **71 card ids and one interned string per
divergent card**.

### Four layers, each hidden behind the one before it

The first attempt was a leaf change alone, and it was **dead code** — nothing could reach it.

| # | layer | what it was doing |
|---|---|---|
| 1 | **the leaf** | `FilterExpr::ExactName` read `card_name_folded` and nothing else, so it answered False for a name only the reversible printing prints. It now consults the divergent record and answers `PrintingDep`, because only the printings whose layout is the divergent one print that name — exactly what `divergent_of` already decides. |
| 2 | **the narrowing** | `narrow_rec`'s ExactName arm goes through a trigram index built from the **card's** collated name, and `templegardentemplegarden` shares no trigram window with `templegarden`. The card was dropped before any leaf ran, which is why a leaf-only fix was unreachable. |
| 3 | **the folded form** | `collate_name`'s parameter is named `folded_lower` and means it — it strips non-alphanumerics and does **not** lowercase. Handing it the display name yields `TempleGardenTempleGarden`, which no needle can equal. |
| 4 | **tightness** | With 1–3 in place the query answered **20** — every printing of Temple Garden. `Narrowed::tight` is a **card-space** claim and stops being true the moment a card is present for a name only *some* of its printings print. |

### What is stored

`CardIndexes` gains **`name_divergent`** — the 71 card ids with such a printing, each with the
folded joined name. Postings rather than a plane for the same reason `legal_divergent` beside it is
a list: 71 of 38,626, cheaper as ids at any density this set could reach. The arm unions them in and
scans **71**, not the corpus.

`DivergentPrinting` gains **`card_name_folded_id`**, interned off the row's own `card_name_folded`
— layer 3. Folding is build-time here; the engine has no runtime accent fold.

Layer 4 is a flag, not a byte: only the **divergent additions** cost tightness. A needle the trigram
index answered by itself is still exact at card level, which keeps `!"Lightning Bolt"` on the cheap
path. `tight_narrow_space` still reports `ExactName` as tight card space and stays sound — it is a
static pre-reject, and the `Not` arm re-reads the flag the arm actually returns.

### The alternative, rejected

Make the arm **decline** when it finds nothing and let the scan do the work. It works, and it sends
every no-match `!"..."` to a full corpus scan for one query form's sake — pinned against by the
existing tight-set tests. 71 ids cost less.

### Why the name rides along with the id

`narrow_rec` runs off `CardIndexes` and the card rows and has **no string table in hand**. Threading
one through it and its ~50 call sites to save 71 short strings is the worse trade. The interned id
is what the per-card **leaf** reads — which does have `strings` and does not have the indexes.

### The consumer table's one correction

`estimator::has_printing_varying_leaf` is listed above under "checked and NOT affected". It was
right about `Layout` and wrong about `ExactName`, and it is the table that fails **silently**. Both
it and `filter::leaf_compares_printing_field` now classify `ExactName` printing-dependent.

### Measured

With `include_extras` and `unique=prints`:

| query | Scryfall | before | after |
|---|---|---|---|
| `!"Temple Garden // Temple Garden"` | 1 | 0 | **1** |
| `!"Mechtitan // Mechtitan"` | 1 | 0 | **1** |
| `!"Adrix and Nev, Twincasters // …"` | 1 | 0 | **1** |
| `!"Temple Garden"` | 20 | 20 | 20 |
| `name:"Temple Garden"` | 20 | 20 | 20 |
| `!"Fire // Ice"` | 9 | 9 | 9 |

### Tests

`exact_name_answers_the_joined_name_on_the_printing_that_prints_it` pins all four layers on one
fixture: the narrow reaching the card **at all** and arriving **loose**, the leaf answering
false/false/true across three printings, `PrintingDep` in card space, and 1 against 3 printings end
to end. Three needles the trigram index answers by itself are asserted **still tight** — the
regression the declining alternative would have caused.

`ARCHIVE_FORMAT_VERSION` 2026081611 → **2026081612**: both `CardIndexes` and `DivergentPrinting`
grow.

### Still open, the other half

`assign_name_ranks` ranks the **card's** collated name, so those 81 printings sort one slot off
under `order=name`. The folded divergent name this commit stores is what a printing-space rank would
be built from. Not carried here.

---

## Watermark: It Is Per Face, and This Stored Only the Front's — In Both Directions

Added to this PR because every prerequisite is already here: `PrintingFace`, `divergent_of`, the
`second_text_field_value` mechanism the `layout:` work introduced, and the `_FACE_OBJECT_FIELDS`
tuple this branch already extended for the face-level `layout`. It is the same shape of finding
against the same structures.

`_FACE_OBJECT_FIELDS` did not carry `watermark`, so a non-front face's value was discarded at
ingest while `_merge_processed_faces` copied face 0's up to `card_watermark`. One omission,
two opposite defects.

**The search index was short.** `Research // Development` (dis/155) is simic on its front face and
izzet on its back, and api.scryfall.com answers it for **both** `wm:simic` and `wm:izzet`; this
answered simic alone.

**The card object was long.** Scryfall sends a top-level `watermark` on 36,437 printings and on
**0** of the 12,098 that have `card_faces` — no layout, no face count and no image count produces
an exception. So every faced printing with a watermark emitted a key Scryfall sends on none of
them. A split card is the shape that hides it: one image, one piece of cardboard, so no layout gate
stands in for "has faces".

### Measured, over the whole 2026-08-16 bulk

| | |
| --- | --- |
| faced printings carrying a watermark (default_cards) | 156 |
| …of those, carrying **two distinct** ones | 19 |
| …carrying three | 0 |
| …whose FRONT face lacks it while a later face has it | 0 |
| faced printings with a top-level `watermark` on Scryfall (all_cards) | **0 of 12,098** |
| unfaced printings with one | 36,437 |
| guilds short by 1–2 rows | all ten |

Sweeping all 76 distinct watermark values against a rebuilt store: **exactly 10 changed**, the ten
Ravnica guilds, +1 or +2 each. Of the 651 faced printings in `all_cards` carrying a face watermark,
**0** now emit a top-level one and all 651 have face watermarks byte-identical to Scryfall's bulk;
a 400-printing sample of the 36,437 unfaced ones is unchanged.

### What moves

- `_FACE_OBJECT_FIELDS` gains `"watermark"`, between `flavor_text` and `artist` — Scryfall's own
  position on every one of the 1,075 face occurrences in the bulk.
- `PrintingFace` gains `card_watermark_id`. That row is where a per-face, per-printing value
  belongs, and it is also the only place it fits: `Printing` has no padding left, so a second id
  there rounds the row across every printing for a value 156 of them carry, while `PrintingFace`
  exists only for the faced ones.
- `indexes.watermarks` holds **every** value a printing carries, deduped per printing
  (`build_watermark_index`, extracted so the differential test calls the same code the commit pass
  does rather than a copy that can drift).
- `second_text_field_value` → `extra_text_field_values`, an unbounded allocation-free iterator;
  `TextExact` and `TextRegex` both route through one `tri_over_values`.
- `ARCHIVE_FORMAT_VERSION` 2026081612 → **2026081613**.

### Why the filter was not the hard part

`narrow_rec` marks the watermark postings **TIGHT** and the compose path calls them **EXACT**, and
nothing downstream re-checks a leaf either one claimed. So the index and `tri` must enumerate the
same set — and a cap on what `tri` can see ("one more value") would be a cap the index does not
share: a third face watermark would have to either drop out of the filter silently or panic the
build. It does neither.

The per-printing **dedupe** is load-bearing on its own: `compose_printing_estimate` reads the
postings LENGTH as the match count, so the 137 faced printings whose two faces *agree* would
inflate it while the bitmap stayed right.

Every other operator **declines** rather than approximating, exactly as it did when the field was
scalar: only `Eq` narrows or composes; the ordering ops and the regex reach the general path where
`tri` is the answer.

### Tests

`watermark_is_per_face_and_every_leaf_still_matches_tri` asserts all four answers — the postings,
the tight narrow set, the compose bitmap and the exact estimate — against one ground truth
(`card_pass` + `residual_matches`, the trivalent walk the materializing plan verifies with), for
every operator `wm:` accepts, over a fixture covering both-faces-disagree, both-faces-agree,
unfaced, faces-all-bare and front-bare-back-set. Reverting the per-face `tri` arm fails it by name
on the first assertion; removing the dedupe fails it on the postings assertion.
`test_a_faced_card_carries_its_watermark_on_the_faces_and_never_at_top_level` pins the card-object
half on both sides of the faced/unfaced split.

### Adaptations

Carried from a downstream port where it was measured first; the two trees have diverged, so three
hunks are not byte-identical and the differences are deliberate:

- that tree's `TextField` has two variants this branch does not (`FullOracleTextLower`, `TypeLine`),
  so the exhaustive match here is the shorter list;
- its card object reaches the top-level gate through a `two_image` / face-owned-key table this
  branch does not have, so the gate here is the direct `key == "watermark" && faces.is_some()`;
- its `#907` regex dialect is not on this branch, so the new test uses `regex::Regex`.


daveycodez:
### Merge order: land this **last** of the five

Full ordering and conflict-resolution rules: https://github.com/jbylund/sylvan_librarian/pull/912#issuecomment-5487092699

Specific to this branch at `5fd256c9`:

- It contains an **older** copy of #927 (`multilingual-store`), which has since moved ahead — take #927's newer commits at the merge, not this branch's copy.
- It holds the **highest `ARCHIVE_FORMAT_VERSION` of the five, 2026082708** (against 2026082707 on #927/#928 and 2026082703 on #912/#922). Whatever order the others land in, this value must be the one that survives; a merge that takes a lower number leaves the merged code building stores that claim a format they do not have.
- `card_object.rs` / `objects.py` here are ids-only, a generation behind #912's, and `card_object_parity.json` does not exist here. After #912 is in `main`: **keep `main`'s copies**, keep the golden file on a modify/delete conflict, do not regenerate it.
- `exact_name_matches` (the `!` operator, this branch) and `name_key_tier` (#912's route-side rule) are the **same two-halves rule on two different surfaces** and both should survive the merge. The reversible/divergent-name arm this PR is actually about is separate from both and is unaffected: a reversible printing has exactly two faces by construction, so its joined key and both face keys stand.

PR #930, opened 2026-08-17
Art Tags: A Card Answers for Every Face It Shows, Not Just Its Front

daveycodez:
**Stacked on #894.** Its commits are in this diff because `dfc-face-merge` is not a branch in this repo; **the last commit is the whole of this change**. Happy to squash it into #894 instead if you'd rather review one thing — I kept it separate because the `prefer_score` consequence at the bottom is a judgement call you should get to make on its own.

## The defect

`card_art_tags` is attached by joining on `magic.cards.illustration_id`:

```python
_sync_card_tags(conn, "illustration_id", "card_art_tags", illustration_id_to_tags)
```

On `main` that is right, and it is right *because* faces are separate rows: a double-faced card's back face is its own row carrying its own `illustration_id`, so a tag on the back art lands on a row and `arttag:` finds the card.

#894 merges the faces into one row and gives it the **front's** `illustration_id` — matching Scryfall's own top-level field, correctly. But then the back's illustration exists nowhere on the row, and a tag that lives only on the back art becomes unreachable by any query.

Measured against **api.scryfall.com**, 2026-08-16:

| query | Scryfall | front-only |
|---|---|---|
| `arttag:snow e:khm` | **75** | 73 |
| `-art:human e:khm t:creature` | **135** | 136 |

The two missing from the first are `Birgi, God of Storytelling // Harnfel, Horn of Bounty` and `Esika, God of the Tree // The Prismatic Bridge` — their snow is on the back face's art. The surplus in the second is `Valki, God of Lies // Tibalt, Cosmic Impostor`: Tibalt is the human, and Tibalt is the back.

## The rule, checked against the corpus rather than assumed

Replaying the 2026-08-16 art-tags dump over the bulk cards, the union predicts Scryfall's `arttag:snow e:khm` answer **set-for-set** — 75 names, no surplus and no shortfall — where the front-only reading predicts 73.

Scope on the same bulk: **9,368** printings show more than one illustration, and **5,491** of them gain at least one tag from a non-front face.

## Shape

**A column, `illustration_ids jsonb`** — every illustration the row shows, front first, deduped. `illustration_id` stays the front's for display.

The merge-side change is one tuple entry:

```python
_FACE_LIST_UNIONS = ("card_types", "card_subtypes", "illustration_ids")
```

because each face row already computes its own single-element list, and `_merge_processed_faces` already unions front-first with dedupe. A split or adventure card — one `illustration_id` on the card, none on the faces — has both face rows inherit the same id and collapse back to one entry, which is the right answer: there is one piece of art to tag.

`jsonb` and not `uuid[]`, for a mechanical reason as much as a stylistic one: `api/db/bulk_upsert.py` resolves a column through `_PG_TYPE`, where `jsonb` extracts with `obj->` and an `ARRAY` column falls through to `text` and fails the insert. Matching `card_types`/`card_subtypes` keeps `bulk_upsert` untouched.

**The sync is keyed on `scryfall_id`, not `illustration_id`.** This is the part I'd most want a second opinion on. An illustration id no longer determines a row's whole value — a double-faced card's tags come from two of them — so no record keyed on one illustration can write the row. It also cannot be batched that way: `_sync_card_tags` batches at 5,000, and a batch of illustration ids can split one card's illustrations across two statements, where a batch of cards cannot.

So `_union_art_tags` resolves the union in Python and `_sync_card_tags` is **unchanged** — still shared byte-for-byte with the oracle path, still diffing and batching, still `IS DISTINCT FROM`-guarded. The costs, stated plainly:

- **Memory: flat.** A single-illustration card references the dump's existing tag dict rather than copying it, so allocation is bounded by those 9,368 multi-illustration printings, not by the corpus.
- **Wire: one record per tagged *card* instead of per tagged *illustration*.** That is the real price. It buys a `scryfall_id` join on the unique index instead of `idx_cards_illustration_id`, and a clear step that is exact per row.

I considered doing the union in SQL over the array instead. It needs either a whole-table `UPDATE` with a per-row aggregate — the shape #910 was about — or a temp table plus an unbatched statement. Neither seemed worth giving up the batching for.

**The migration backfills** from `raw_card_blob->'card_faces'`, which on an already-imported row is the only record of the back's illustration. Without it the column is empty until the next card import and `_sync_card_tags` clears every card's art tags in the meantime. The backfill is exercised against real imported rows in `test_upsert_cards.py`, so it cannot drift from what `preprocess_card` writes.

## `prefer_score` reads the union too, deliberately

`art_style` in `backfill_prefer_scores.sql` reads `card_art_tags`, so this moves representative selection. I measured it rather than guessing: **exactly three** printings corpus-wide flip `art_style` under the union.

- `Tribute to Horobi // Echo of Death's Wail` neo/356, English and German — back art is `anime`
- `Thaumatic Compass // Spires of Orazca` pxtc/249 — back art is `line-art`

I fed the union to `prefer_score` rather than keeping a front-only reading for it, and that is a choice, not a consequence. The argument for it: `art_style` asks what art a printing shows, and a printing shows all of it — a card whose back is anime *is* partly anime. The argument against: two readings would protect three rows from a demotion #720's evidence didn't rule on.

Three rows is small enough that I'd take either answer; say the word and it's a one-line split. Note that `artwork_printings` in the same file still groups on the front `illustration_id` and is untouched — that is a different question (which printings are the same artwork), and I did not want to move it here.

## Equivalence with the Cloudflare port

I carry the same fix in my Workers port of this project, where the import is Rust over a flat store rather than Python over Postgres. The implementations are unrelated — there the union is `transform::art_tags_of`, computed in-process at build time over the row's `card_faces`, sorted and deduped because the store's tag lists are sorted slug text; here it is a `dict.update()` fold over a maintained column, because a jsonb object has no order to preserve.

They agree on **behaviour**, which is what matters and what is measurable: both take the row's own illustration plus each face's, front first and deduped, skipping a face that carries no art of its own; both feed the same union to the representative score. Both produce 75 for `arttag:snow e:khm` and 135 for `-art:human e:khm t:creature`, against 73 and 136 for the front-only reading.

## Tests

Against real Postgres, not mocks (`api/tests/test_tagging_integration.py`, `api/tests/test_upsert_cards.py`):

- a back-face tagging lands on the merged card — this fails on the front-only sync, verified by reverting the keying and watching it go red
- the front-face tagging still lands, unchanged
- a dump with no taggings clears the row, so the diff-and-clear path stays exact under the new key
- a merged row stores both faces' illustrations, a single-faced row stores its one
- the migration's backfill lands where `preprocess_card` would, on rows written by the real import path

And as units (`api/tests/test_card_processing.py`, `api/tests/test_tagging_integration.py`): front-first ordering, a split card's faces collapsing to one illustration, a face with no art contributing nothing, an untagged illustration omitted rather than mapped to `{}`, and the no-copy property of the single-illustration path.

`python -m pytest --ignore=api/tests/test_integration_testcontainers.py` on this branch: **2,551 passed, 19 xfailed**. That file errors 22 times for me, identically on `dfc-face-merge` without this commit — its class-scoped container binds a fixed host port and collides with the session container locally, so I've left it to CI. `ruff check` and `ruff format --check` clean.

### DIFF
--- a/api/card_processing.py
+++ b/api/card_processing.py
@@ -208,10 +208,19 @@ def preprocess_card(card: dict[str, Any]) -> list[dict[str, Any]]:  # noqa: PLR0
         if not face_rows:
             return []
         merged_row = _merge_processed_faces(face_rows)
-        # The blob is the front face's dict, so every existing top-level read (image_uris, lang,
-        # finishes, ...) keeps meaning "the front". Re-attach the raw faces for consumers that
-        # need per-face data — the back-face image sync in scripts/copy_images_to_s3.py.
-        merged_row["raw_card_blob"]["card_faces"] = card_faces
+        # The blob is the card-level object with its faces re-attached — what Scryfall sent, not a
+        # face promoted to look like a card. Every searchable field is already merged onto the row
+        # above, so the blob has no derivation left to do, and keeping it verbatim is what makes it
+        # answerable: a card object cannot be rebuilt from a face (`card_faces` is gone, `name` and
+        # `type_line` are the front's, and which fields a real card carries at top level varies by
+        # layout — a split card has `mana_cost` and `image_uris` there, a transform card does not).
+        #
+        # The one consumer that read a *face* field off the blob is `image_uris`, which for a
+        # transform card now lives only under `card_faces`; every reader coalesces to
+        # `card_faces->0` (scripts/copy_images_to_s3.py, scripts/prefer_weights.py). Everything else
+        # read from the blob — lang, set_type, games, finishes, frame_effects, image_status,
+        # reserved, game_changer — is card-level and identical either way.
+        merged_row["raw_card_blob"] = copy.deepcopy(card) | {"card_faces": card_faces}
         return [merged_row]
 
     # Single face case - set defaults
--- a/api/card_processing.py
+++ b/api/card_processing.py
@@ -94,11 +94,66 @@ def extract_collector_number_int(collector_number: str | int | float | None) -> 
     return None  # Field will be null by default
 
 
+# Face-merge policy for multi-face cards (#400, #873). Scryfall AND's search predicates at the
+# CARD level, each satisfiable by any face — measured against api.scryfall.com 2026-08-08:
+# `t:sorcery t:land` returns the MDFC lands (no single face is both), o: conjunctions match
+# across faces (Ral, Monsoon Mage), and `c:b` matches Westvale Abbey's back-face-only color.
+# One row per printing carrying any-face unions reproduces those semantics directly; one row
+# per face would instead break every cross-face conjunction (no face-row satisfies both terms)
+# on top of colliding on the scryfall_id primary key, which is how the back face silently won
+# until now. Front-face scalars (cmc, mana cost, illustration, image, prices) match Scryfall's
+# own top-level fields, verified on its card objects.
+_FACE_LIST_UNIONS = ("card_types", "card_subtypes")
+_FACE_FLAG_UNIONS = ("card_colors", "card_keywords", "produced_mana")
+_FACE_JOINED_TEXTS = ("oracle_text", "flavor_text", "type_line")
+# Copied per GROUP from the first face that has any of the group, so the numeric columns and
+# their _text twins always describe the same face (the schema's check constraints couple them).
+_FACE_STAT_GROUPS = (
+    ("creature_power", "creature_toughness", "creature_power_text", "creature_toughness_text"),
+    ("planeswalker_loyalty", "planeswalker_loyalty_text"),
+)
+# Joins face texts. "\n" so substring/regex matches cannot span faces in practice (`.` does not
+# cross newlines), "//" because that is the face separator Scryfall itself renders.
+_FACE_TEXT_SEPARATOR = "\n//\n"
+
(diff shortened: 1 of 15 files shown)


daveycodez:
Folded into #894 (commit a9bfc56, fast-forwarded onto `dfc-face-merge`) and closing this.

Splitting it out was the wrong call. The defect this fixes does not exist on `main` — there, faces are separate rows, so a double-faced card's back face carries its own `illustration_id` and a back-art tag lands on a row that `arttag:` can reach. It is #894's face merge that gives the row the front's `illustration_id` and makes the back's art unreachable. #894 has not merged, so keeping these apart would have asked a reviewer to merge a regression and its repair as two units, both of ours, neither landed. One PR that was always correct is strictly better than that.

Everything is unchanged in the fold — the `illustration_ids jsonb` column and its migration, the `_FACE_LIST_UNIONS` entry, the `scryfall_id` keying with `_sync_card_tags` untouched, and the tests including the negative control that turns the back-face test red when the keying is reverted. #894's description now carries the measurements (`arttag:snow e:khm` = 75, `-art:human e:khm t:creature` = 135, the dump replay agreeing set-for-set) and the three `prefer_score` printings that flip, with the one-line split still on the table if you'd rather have it.

PR #933, opened 2026-08-20
Parsing: Stop Accepting Mana Symbols No Real Cost Ever Has

jbylund:
Follow-up from review of #909.

## Two related gaps in `is_valid_mana_symbol`

**Compound symbols were validated side-by-side, not as a whole.** Splitting a
symbol on `/` and checking each side independently against the atom set means
any combination of individually-legal sides passes: `{W/W}`, `{2/2}`, `{C/C}`,
`{W/U/B}`, `{C/P}`. Confirmed these reach production SQL via `hand_parser`
(the live parser per `parsing_f.py`) and compile fine — they just can never
match a row, so the query silently returns zero results forever instead of
400ing the way a genuinely-nonexistent symbol like `{Q}` already does.

Fixed with a flat set of every valid 2- or 3-part shape, compared as the
exact tuple `symbol.split("/")` produces:

- **Hybrid** (`{W/U}`) and **colourless-hybrid** (`{C/W}`) sides can appear in
  either order — a user typing a real symbol from memory has no reason to
  know Scryfall prints `{W/U}` rather than `{U/W}`, and the two name the same
  symbol — generated as permutations of the six symbols `{W,U,B,R,G,C}`.
- **Generic-hybrid** and **phyrexian** don't move: every printing has generic
  first (`{2/W}`, never `{W/2}`) and phyrexian last (`{W/P}`, never `{P/W}`),
  so those two stay one-directional.
- **Hybrid-phyrexian** (`{R/G/P}`) is generated for all 10 colour pairs, not
  just the 4 the card corpus happens to have printed — the shape itself
  (two distinct colours + phyrexian) is what makes a request valid, not
  whether a specific pair has shipped on a card yet.
- A side never repeats (`itertools.permutations` never repeats an element in
  one draw, so `{W/W}`/`{2/2}`/`{C/C}` are excluded structurally, no separate
  check needed).
- Never valid, by rule rather than corpus: colourless combined with generic
  or phyrexian (`{C/P}`, `{2/C}`) — Phyrexian mana is only ever a colour's
  (comprehensive rules 107.4d), and generic-hybrid is only ever generic+colour.

**`∞` and half mana were accepted despite being un-searchable by
construction.** `_ATOMS` included `'∞'` (Gleemax's whole cost, from Unglued)
and half mana (`'H'` + a colour, from Unhinged/Unstable) as valid symbols.
Both are real printed symbols, but only on cards from `set_type == "funny"`
sets, which `preprocess_card` filters out of the corpus entirely (`api/card_processing.py:108`).
So a mana cost search for either one can never match a row, same failure
mode as the compound-symbol gap above — just for two specific symbols
instead of a shape. Moved them from "valid" to `_REAL_BUT_NOT_A_COST`,
alongside `{T}`/`{Q}`/`{CHAOS}` and the other real-but-uncostable symbols.

## Verification

- `api/parsing/tests/test_mana_symbols.py`, `test_spans.py`,
  `test_regex_patterns.py`, `test_parser_parity.py`: updated/passing.
- Full `api/parsing/tests/` suite: 2057 passed, 19 xfailed.
- `ruff check` / `ruff format --check`: clean.
- Manually verified both parsers agree, and every position/order rule
  behaves as designed: `{w/u}`/`{u/w}`, `{c/w}`/`{w/c}`, `{r/g/p}`/`{g/r/p}`,
  `{u/b/p}`/`{b/u/p}` all valid; `{w/2}`, `{p/w}`, `{w/w}`, `{2/2}`, `{c/c}`,
  `{c/p}`, `{2/p}`, `{w/u/b}` all rejected.

### DIFF
--- a/api/parsing/mana_symbols.py
+++ b/api/parsing/mana_symbols.py
@@ -32,34 +32,33 @@ _ATOMS = _COLORS | frozenset("CSXYZP")
 # The generic side of generic-hybrid mana is always specifically '2' ({2/W}, never {1/W} or {3/W}).
 _GENERIC_HYBRID_VALUE = "2"
 
-_TWO_PARTS = 2
-_THREE_PARTS = 3
+# Colourless joins colour on the "which side does this go on" ambiguity a hybrid pair has — a user
+# typing {C/W} from memory has as little reason to know it isn't printed {W/C} as they do for two
+# colours — so both share one permutation-generated set of ordered pairs. Generic and phyrexian don't:
+# every printing has the generic side first and the phyrexian side last, so those two are one-directional.
+_HYBRID_ATOMS = _COLORS | frozenset("C")
 
-# A multi-part symbol's sides can appear in either order — a user typing from memory has no reason to
-# know Scryfall prints hybrid as '{W/U}' rather than '{U/W}', and the two name the same symbol — so
-# each shape is stored as a frozenset of its sides, not a tuple. What a *set* can't capture is a side
-# repeating itself: '{W/W}' and '{2/2}' aren't symbols even though 'W' and '2' are each legal alone, so
-# duplicate-side combinations are rejected before this table is ever consulted.
-_TWO_PART_SHAPES = frozenset(
-    frozenset(pair)
-    for pair in (
-        *itertools.combinations(_COLORS, 2),  # hybrid: any two colours, e.g. {W/U}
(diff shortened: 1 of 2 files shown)


jbylund:
This should not be curious based, and we should generate all possibilities of len 1-3 and just do an in check.

PR #957, opened 2026-08-21
Parsing: Make Bare Mana Characters Agree With Their Braced Form

### LINKED ISSUE #954: Parsing: pyparsing Silently Drops Bare Mana Characters It Doesn't Recognize
Confirmed still live on current main (post-#909): `mana:s`, `mana:snow`, `mana:p`, and `mana:hello` all still silently produce wrong or empty results via the pyparsing parser, even though the hand parser correctly rejects every one of them.

## Confirmed reachable, right now

```python
>>> from api.parsing.pyparsing_based import parse_search_query
>>> from api.parsing import generate_sql_query, parsing_f
>>> generate_sql_query(parsing_f.parse_scryfall_query("mana:s"))
ValueError: Failed to parse query: "mana:s"          # hand parser: correct
>>> generate_sql_query(parse_search_query("mana:s"))
(..., {'p_dict_e30': {}, 'p_int_MA': 0})              # pyparsing: {} matches every card with a cost
>>> generate_sql_query(parse_search_query("mana:snow"))
(..., {'p_dict_...': {'W': [1]}, 'p_int_MQ': 1})      # pyparsing: silently answers mana:w instead
```

## Root cause

Two independent gaps compound:

1. `pyparsing_based.py`'s mana tokenizer regex (`simple_mana_symbol = Regex(r"[0-9WUBRGCXYZwubrgcxyz]")`
   in two places) accepts `S` as a mana-token character at the grammar level, so `mana:snow` becomes a
   `ManaValueNode` at all — on the hand-parser side, the lexer's own bare-character alphabet doesn't
   include `S`, so the query never becomes a mana value in the first place.
2. Once a `ManaValueNode` exists, nothing on the pyparsing path calls `first_invalid_mana_symbol` to
   validate it the way `hand_parser.parse_mana_value` does — so `mana_cost_str_to_dict`
   (`api/parsing/card_query_nodes.py`) receives the raw, unvalidated string. That function's own
   character loop only recognizes `WUBRGCX`, silently dropping anything else: `S` in `"snow"` vanishes,
   leaving `{'W': [1]}` from the trailing `w`; every character in `"hello"`/`"s"`/`"p"` vanishes,
   leaving `{}` — a subset of every cost, so the query matches every card that has one at all.

## Prior art — do not just reopen or rebase #915

[#915](https://github.com/jbylund/sylvan_librarian/pull/915) diagnosed this exact bug in detail
(verified row counts, the "five copies of what is a cost symbol" table) and proposed unifying every
copy under one shared, public `MANA_COST_ATOMS` constant. Closed rather than rebased: its branch
predates #941 (merged), and its proposed `MANA_COST_ATOMS` value — `_COLORS | frozenset("CSXYZP∞")` —
still includes bare `P`, `Y`, `Z` as valid atoms. #941 established that none of those three ever
appears unpaired in a real (non-funny) mana cost and made `is_valid_mana_symbol("P")` etc. return
`False`; reusing #915's constant as-is would silently regress that fix back in, since the same shared
constant would flow into the bare-character validation this issue is about too.

A fresh fix needs to unify the vocabulary (closing this issue's gap) while keeping #941's exclusions
(not regressing that one) — read both before starting.

## Reference

#915's own diagnosis is still a useful map: five separate copies of "what is a cost symbol" existed —
`mana_cost_str_to_dict`, `calculate_cmc`, and `pyparsing_based`'s two `simple_mana_symbol` patterns
each read their own hardcoded charset (all incomplete in different, overlapping ways), while
`mana_symbols.py`'s (post-#941) `_ATOMS` is the one that's actually correct today.

jbylund:
## Summary

Closes #954.

- `mana:s` now means the same as `mana:{s}` on both parsers. `S` (snow) was missing from `BARE_MANA_ATOMS` in `card_query_nodes.py` — the single alphabet `mana_cost_str_to_dict`, `calculate_cmc`, and (via `mana_symbols.py`) `first_invalid_mana_symbol` all share — even though `mana_symbols.py`'s braced-symbol validator already accepted `{s}`. Adding `S` there fixes bare-vs-braced parity for both parsers at once, and `calculate_cmc`'s bare-character handling now counts a bare `S` toward cmc the same way it already counted `{S}`.
- `mana:snow`, `mana:p`, and `mana:hello` now 400 on the pyparsing path the same way `mana:{s}{n}{o}{w}` already did, instead of silently falling through to a plain string comparison and matching the wrong cards. `pyparsing_based.py`'s `simple_mana_symbol` tokenizer regex was widened from an enumerated letter subset to any letter/digit — which characters are *valid* is `first_invalid_mana_symbol`'s job, not the tokenizer's; a narrow tokenizer charset just means some bare values never reach that validator at all. The `string_value_word` fallback was also dropped from the mana condition's grammar entirely, so a mana/devotion condition's rhs is now always a validated `ManaValueNode`, never an unchecked `StringValueNode` — mirroring `hand_parser.parse_mana_value`'s shape directly.

See `docs/issues/done/00954-pyparsing-bare-mana-charset-drop.md` for the full diagnosis and fix writeup.

## Test plan
- [x] `make test-unit` — 2978 passed, 19 xfailed (unchanged known-divergence xfails)
- [x] `make test-integration` — 22 passed
- [x] `make lint` — clean
- [x] New coverage: `test_bare_mana_character_parity` (parser parity for `mana:s`/`snow`/`p`/`hello`), `test_full_sql_translation_jsonb_colors` cases for `mana:s` vs `mana:{S}`, updated `test_mana_symbols.py` cases for bare `S`


jbylund:
Opened against the wrong branch (a shell cwd reset put this on close-out-909-review-issues instead of mana-bare-atom-parity). Closing and reopening correctly.

PR #961, opened 2026-08-21
Admin: Shrink the Parent/Child Shared Surface

jbylund:
## Summary

Follow-up to #794's review: shrinks the parent/child shared surface that PR documents (five methods, four handles) down to two methods and three handles, plus AdminResource getting its own injected connection pool.

- Inline the existence check in `import_card_by_name` instead of going through the parent's cached, EXPLAIN-capable `_run_query` for a plain lookup.
- Extract `_set_statement_timeout` to `db_utils` — it never touched `self`.
- Extract `_serve_static_file` to `page_rendering` — fully stateless, and `page_rendering` already owns `STATIC_DIR`.
- Give `AdminResource` its own `psycopg_pool.ConnectionPool` instead of reaching into the parent's.
- Have `APIResource.__init__` build both pools and inject the admin one into `AdminResource`'s constructor, instead of `AdminResource` calling `db_utils.make_pool()` itself — mirrors the existing injectable handles (`import_guard`, `last_import_time`, ...) and lets tests hand in a mock at construction time instead of opening a real pool just to close it.

What's left in the shared surface (`_reload_engine`, `_setup_complete`, `_cache_generation`, `_last_import_time`, `_setup_complete_cache`) is the genuinely irreducible part: `multiprocessing` primitives that are the actual cross-worker-process signal for "the corpus changed" / "check whether setup finished," not incidental references.

## Test plan

- [x] `make test-unit` equivalent (full suite minus the docker integration file): 2994 passed
- [x] `test_integration_testcontainers.py` (docker): 22 passed
- [x] Full suite together: 3016 passed
- [x] `ruff check` / `ruff format --check` on all changed files

### DIFF
--- a/api/admin_resource.py
+++ b/api/admin_resource.py
@@ -9,8 +9,8 @@ public method became a route, and the only lever was a leading underscore, which
 method's Python visibility. Mounting a separate resource replaces that lever with a boundary, and
 keeps the honest names.
 
-The child holds a reference to its parent for the small surface they genuinely share — four methods
-(`_reload_engine`, `_serve_static_file`, `_set_statement_timeout`, `_setup_complete`)
+The child holds a reference to its parent for the small surface they genuinely share — three methods
+(`_reload_engine`, `_serve_static_file`, `_setup_complete`)
 and four handles (`_conn_pool`, `_cache_generation`, `_last_import_time`, `_setup_complete_cache`).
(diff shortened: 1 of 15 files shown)


jbylund:
Superseded by #963 (same reasoning as #794 — verified byte-identical diff, recreated as one PR against `main`). Closing without merging.

PR #962, opened 2026-08-21
Introduce AppContext, Remove AdminResource's Parent Reference

jbylund:
## Summary

Follow-up to #961: introduces `AppContext`, the object holding the state genuinely shared between peer resources rather than owned by one — connection pools, the query engine, and the cross-worker cache/import signals. `AdminResource` no longer holds a `_parent` reference to `APIResource` at all.

- **`api/app_context.py`** (new): `AppContext` holds `reader_pool`, `writer_pool`, `engine`, `engine_reload_guard`, `cache_generation`, `last_import_time`, plus the `setup_complete`/`reload_engine`/`bump_cache_generation`/`invalidate_setup_complete` logic that operates on them (moved from `APIResource`).
- **`APIResource.__init__`** now takes `app_context`/`admin_context` instead of five granular params (`conn_pool`, `admin_conn_pool`, `cache_generation`, `engine_reload_guard`, `import_guard`, `schema_setup_event`, `last_import_time`).
- **`AdminResource.__init__`** drops `parent`/`conn_pool`/`import_guard`/`schema_setup_event`, takes `app_context`/`admin_context` instead. Every `self._parent.X` access is gone.
- **`AdminContext`** (new, defined in `admin_resource.py`): `import_guard`/`schema_setup_event`, kept separate from `AppContext` since they're cross-worker-shared only *within* AdminResource's own copies — nothing on the search side touches them. `APIResource.__init__` forwards it opaquely without needing to know its shape.
- **`api_worker.py`**: `get_api()` now builds both context objects post-fork from the same five primitives it already received; its own signature and `entrypoint.py` are unchanged.
- Every test constructing `APIResource(...)`/`AdminResource(...)` directly, plus `scripts/bench_random.py`, updated for the new constructor shape. `api/tests/support.py`'s `mock_conn_pool_kwargs()` is replaced with `mock_app_context()`.

One correctness bug surfaced during the migration and is fixed here: three tests in `test_parsing_errors.py` that patch a pool and call `reload_engine()` were patching `reader_pool`, which `reload_engine` no longer reads from (it deliberately reads via `writer_pool` — see the docstring on `AppContext.reload_engine`) — fixed to patch `writer_pool`.

Two follow-up commits address review discussion:
- `reload_engine`'s docstring cited an unmeasured "can run for minutes" claim justifying the `writer_pool` choice. Measured against a real ~98k-card corpus: ~3 seconds end-to-end, matching `docs/issues/done/00505-engine-incremental-loading.md`'s "seconds-long" characterization. Docstring corrected to cite the number and note the in-process engine serves the large majority of search traffic without touching `reader_pool` at all — so the choice stands, but was never as consequential as originally implied.
- Added `api/tests/test_app_context_cross_process.py`: every other `AppContext` test builds one instance and calls its methods directly, which never exercises the actual cross-worker claim — that `cache_generation`/`last_import_time`/`engine_reload_guard` stay genuinely shared across a fork. These spawn a real child process (`multiprocessing.get_context("fork")`, explicit so behavior matches production regardless of host OS) and prove a write in one process is visible from an independent `AppContext` built from the same primitive in another.

## Test plan

- [x] Full suite (unit + docker integration): 3032 passed, 7 deselected (pre-existing stale-engine failures unrelated to this change), 19 xfailed
- [x] `ruff check` / `ruff format --check` on every touched file
- [x] Manual sanity check of `ApiWorker.get_api()`'s actual production wiring path (not exercised by any existing test) — builds the Falcon app successfully with mocked pools
- [x] Real measured timing of `reload_engine(force=True)` against a populated dev DB (~98k cards): ~3 seconds

### DIFF
--- a/api/admin_resource.py
+++ b/api/admin_resource.py
@@ -9,20 +9,16 @@ public method became a route, and the only lever was a leading underscore, which
 method's Python visibility. Mounting a separate resource replaces that lever with a boundary, and
 keeps the honest names.
 
-The child holds a reference to its parent for the small surface they genuinely share — two methods
-(`_reload_engine`, `_setup_complete`) and three handles (`_cache_generation`, `_last_import_time`,
-`_setup_complete_cache`). All three of those handles are `multiprocessing` primitives: the actual
-mechanism by which one worker process tells every other worker "the corpus changed" or "check
-whether setup finished," not incidental references. That is deliberately not a decoupling: the
-boundary here is about routing, and pretending otherwise would mean an `AppContext` refactor that
-the routing fix does not need.
+`AdminResource` no longer holds a reference to `APIResource` at all. What it used to reach a parent
+for — `reload_engine`, `setup_complete`, `cache_generation`, `last_import_time`,
+`invalidate_setup_complete`, the connection pool — all live on the `AppContext` both resources take a
+reference to at construction (see `api/app_context.py`). Neither resource owns that state; both are
+peers that reach into a neutral shared object instead of into each other.
 
-The child keeps its own `_conn_pool` rather than sharing the parent's: nothing here needs the
-parent's query-result cache or EXPLAIN plumbing, only a connection, so a second pool removes that
-coupling for the price of a second `psycopg_pool.ConnectionPool` (and its own connections) per
-worker process. `APIResource.__init__` builds both pools and hands this one over at construction,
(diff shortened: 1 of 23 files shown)


jbylund:
Superseded by #963 (same reasoning as #794 — verified byte-identical diff, recreated as one PR against `main`). Closing without merging.

PR #1008, opened 2026-08-23
Engine: Fix Routing-Regret Sweep's Top Cost-Model Bugs

jbylund:
Stacked on #1007. Five commits, found and fixed via a uniform-sampler routing-regret sweep (`scripts/bench_regret_matrix.py` / `scripts/bench_cost_model_agreement.py`) rather than a specific bug report -- each one closes a gap the sweep surfaced.

## Commits

**`Exact Walk-Length Checkpoints for otag:/keyword:/t: Ordered by EDHREC`** + **`Price a Card-Space Collection Leaf's Compose Build Separately From a Range's Scatter`** -- `PrintingCompose`'s Perm-branch walk-length estimate (`printings_walked`) is blind to EDHREC-order clumping for card-space collection leaves. Adds 5 exact sampled checkpoints (cumulative printings examined at card-permutation positions 1/100/200/400/800) per (value, direction), interpolated at query time, plus a second fix: the collection-leaf compose build was priced with a range's cheap scatter rate instead of the pricier card-broadcast rate its actual build operation (`ids_of` + `broadcast_card_ids_to_printings`) uses.

**`Fix scan_units Argument Swap in Cost-Model Gold/Refit/Probe Tests`** -- `scan_units()`'s last argument is the true total `n_cards` (matches the real router's call site), but three test harnesses (`plan_cost_model_matches_gold`, `plan_cost_refit`, `printing_range_route_probe`) passed `eval_domain` instead, wildly inflating the model's own self-check on narrow queries. `cmc>=15`/`o:annihilator`'s reported "120x misroutes" were entirely this bug -- confirmed the real router already routes them correctly via `explain_analyze` against the real corpus before touching anything else. `plan_cost_model_matches_gold`: 90.9% -> 98.9%.

**`Price GatheredScan's Artwork-Mode Dedupe Check`** -- the regret sweep's dominant misroute cluster among 100us+ queries was `StreamedSelect`<->`GatheredScan` confusion. Half of it: `exec_gathered_scan`'s `Mode::Artwork` branch runs the same `group_best`/`touched` dedupe work `StreamedSelect` charges via `artwork_seen_cards`, but `GatheredScan`'s cost formula charged nothing for it, so it read as a near-exact tie against `StreamedSelect` while measuring 10-16% slower in every observed case. Fixed by adding a new `artwork_seen_printings` feature and a `GATHER_ARTWORK_PER_PRINTING_NS` cost term to `GatheredScan`'s formula -- rate fit via a dedicated kernel test (`gather_artwork_kernel_costs`) rather than an end-to-end A/B or a pooled regression, since both of those proved unusable: the first dominated by the two plans' very different `matches_pushed`, the second by multicollinearity between `cards_visited`/`printing_span`/`matches_pushed` in natural query data. Verified against all three concrete misroutes the sweep found: all three now correctly route to `StreamedSelect`.

**`Narrow layout: (is:split/dfc/meld/...) Instead of Scanning Everything`** -- the other half of the misroute cluster, and the bigger find: `layout:` (what `is:split`/`is:dfc`/`is:meld`/... rewrite to) had no candidate-narrowing arm at all, unlike `border`/`set_code`/`watermark`, so any `layout:`-shaped query fell back to "assume ~everything matches" -- `is:split eur>0.07` estimated 31,724 of 31,724 candidate cards against a real 132. Gave `layout` the same `HybridTagIndex` treatment `subtypes`/`keywords`/`oracle_tags` already have. Bigger than a routing fix: these queries now actually narrow, so they run 30-60x faster end to end (`is:split eur>0.07` ~200us -> ~3-5us), not just route to the right plan.

**`Fix scan_units' Card-Mode Overcharge Under Default Prefer`** -- percentile breakdown of `scan_units/printings_examined` by (plan, unique, residual-present) surfaced a clean, constant 3.08x error across the ENTIRE distribution (flat p10-p90) for `GatheredScan`/card/all-match -- 3.08 is exactly `n_printings/n_cards`. `card_match_count`/`push_card_matches`'s `Mode::Card` arm under `Prefer::Default` breaks at the first printing whenever `all_match` is true, so `printings_examined` is always exactly 1 per card, never the corpus-average span `scan_units` estimated. `GatheredScan` reads `scan_units` unconditionally (no tier gate, unlike `StreamedSelect`'s own scan term), so the error rode straight into every such query. Verified: the ratio is now flat 1.00 for this population (was flat 3.00-3.08); median measured/predicted for the no-residual population moved 0.814 -> 0.864, within-25% 53% -> 63%.

Plus two small tooling fixes found and needed along the way (`bench_cost_model_agreement.py`'s missing `prepare_candidates` netting, `fit_cost_model.py`'s stale `artwork_seen_printings` mirror), and two documentation-only commits recording investigations that were tried and deliberately NOT shipped (a card-mode `matches` estimator bug with no cheap fix, and a P3+P4 joint refit that improved the gold test but regressed total routing regret 12% via `PrintingCompose` -- see `docs/issues/local-engine-card-residual-pass-rate.md` and `docs/issues/local-engine-p3-p4-joint-refit-vs-compose.md`). `card_engine/src/cost.rs`'s rate constants are unchanged from before those two investigations; only the `scan_units` feature fix above is a routing-relevant code change from them.

## Verification

- `cargo test --release`: 156 passed, 0 failed, at every commit in the stack
- `cargo clippy --release --all-targets`: clean (one pre-existing unrelated dead-code warning)
- `plan_cost_model_matches_gold` (rebuilt `real.store` after each archive-format bump): 90.9% -> 98.9% -> 97.7% (layout fix, one new unrelated-borderline miss) -> 98.9% (scan_units card-mode fix)
- Every fix verified against the real corpus via `explain_analyze` on the specific misrouted queries the regret sweep surfaced, not just the aggregate metrics
- The two reverted/unshipped investigations were verified NOT to help via the same regret-matrix instrument, and documented rather than silently dropped

## Test plan
- [x] `cargo test --release -p card_engine`
- [x] `cargo clippy --release --all-targets -p card_engine`
- [x] `plan_cost_model_matches_gold -- --ignored` against a freshly rebuilt `real.store`
- [x] Concrete misrouted queries re-verified against the real corpus after each fix
- [x] Full uniform-sampler regret-matrix sweep before and after, checked for cross-plan regressions, not just the targeted transition

### DIFF
--- a/card_engine/src/bench_gather_loop.rs
+++ b/card_engine/src/bench_gather_loop.rs
@@ -153,6 +153,11 @@
 //! support for that ordering -- P3's dispatch median is 32 us against P4's 4 us, and its finish phase is
 //! 12% of all measured ns at mean |log| 2.06.
 //!
+//! **That "applying both arms together" attempt happened (2026-08-23, `bench_streamed_loop`'s own
+//! module docs), after a real feature fix, and still regressed -- through P3/P4-vs-`PrintingCompose`
+//! distortion, not P3-vs-P4 compensating error. See the sixth row of that file's history table before
+//! trying this again: a joint refit needs every plan in the same argmin refit together, not just P3+P4.
+//!
 //! One further limit on all of these: `LIMIT` is fixed, so nothing here constrains how the rates
 //! behave as the page grows.
 //!
--- a/card_engine/src/bench_streamed_loop.rs
+++ b/card_engine/src/bench_streamed_loop.rs
@@ -102,6 +102,54 @@
 //!     P4 reparameterised                        +40%
 //!     P3 and P4 both refit                      +30%
 //!     ... plus symmetric residual floors        +90%
+//!     P3+P4 refit AGAIN after fixing scan_units +54% (mean regret 0.50 -> 0.56 us/query)
+//!
+//! **The sixth attempt (2026-08-23) is not a repeat of the first four.** It came after fixing a real
+//! feature bug (`scan_units`'s card-mode overcharge, `local-engine-card-residual-pass-rate.md`'s
+//! companion) and refitting P3+P4 jointly on the corrected data -- the exact prerequisite this file's
+//! own conclusion asked for. It still regressed, but through a DIFFERENT mechanism than the first five:
+//! not P3-vs-P4 compensating error (fixing the feature and refitting both arms together handles that),
+//! but P3/P4-vs-`PrintingCompose` distortion. `PrintingCompose` was deliberately left unrefit (its own
+//! `Perm` arm is missing a `cards_visited` feature -- see `local-engine-compose-build-rates.md`), so
+//! making P3/P4 cheaper without touching it made them win against `PrintingCompose` in cases where
+//! `PrintingCompose` was actually correct: `PrintingCompose -> StreamedSelect` went from a minor cell to
+//! 309 queries at 99% miss rate and 32% of all lost time, and the `printing_compose` acquire branch's
+//! total share jumped to 72%. Reverted. The lesson generalizes past this one pair: **a joint refit is
+//! only as joint as every plan that competes in the same argmin, not just the two you fixed a feature
+//! for.** `PrintingCompose` needs its own feature fix (the `cards_visited` estimator) before ANY of its
+//! rates -- or any rate for a plan that regularly competes against it -- can be safely moved again.
+//! Full mechanism and revert data:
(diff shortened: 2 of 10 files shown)


jbylund:
Superseded — fully split into #1014, #1015, #1016, #1018, #1019 (all merged) plus #1017 (still open, carrying the remaining walk-length-checkpoints commit). Nothing unique left here.

PR #1017, opened 2026-08-24
Engine: Exact Walk-Length Checkpoints for otag:/keyword:/t: Ordered by EDHREC

jbylund:
Split from #1008 into an individually-reviewable chunk — see that PR's description for the full context this arose from. Fully independent of the other splits.

`docs/issues/local-engine-compose-paging-cost-based.md` leaves the Perm/OrderbyWalk walk-length estimate ("how many printings does the walk examine to fill a page") as a single global rate, and says plainly that "how matches clump along a sort order is not something a density ratio can see, and no constant will fix that." `otag:triggered-ability` measured exactly that: the formula predicted 602 printings walked against a real 1,942 — a 3.2x miss a ~10% printings-per-matching-card skew cannot explain, so the error is EDHREC-order clumping specific to this value, not a rate problem.

**`WalkCheckpoints`** replaces the single rate with five exact anchors per (value, direction): the permutation position at which cumulative PRINTINGS from matching cards reaches 1/100/200/400/800 (not card count — a card-level match can contribute many printings at once, so the threshold has to be printing-weighted). Interpolated at query time between whichever two anchors bracket `offset+limit`. 20 bytes per (value, direction), built once at store time from data already computed for other purposes.

Scoped to `unique=printing` ordered by EDHREC, `subtypes`/`keywords`/`oracle_tags` only: the dominant orderby (40% of `REALISTIC_ORDERBY_WEIGHTS`) and exactly the fields producing every `otag:`-shaped mis-route measured this session. `art_tags`/`is_tags` are printing-space (no card permutation to build against); card/artwork mode and the other 7 orderby columns are unattempted.

Verified against the real corpus: `printings_walked` now predicts 1,998 against a realized 1,942 (2.9% off, down from the old formula's 602 — a 3.2x miss). The mechanism is real: only 50 cards need visiting to accumulate 1,942 printings, because EDHREC's most popular cards are disproportionately old, heavily-reprinted staples (~39 printings/card among the first 50, against a 3.1 corpus average) — genuine clumping, not noise.

**Second commit, found while validating the first**: `compose_printing_estimate`'s `CollectionCmp` arm charges a card-space collection leaf's build at `COMPOSE_SCATTER_PER_PRINTING_NS`, whose own doc comment says it is calibrated for a range's cheap contiguous scatter — but a card-space leaf's actual build (`ids_of` + `broadcast_card_ids_to_printings`) needs a card cursor that rate assumes away. Backed out a real rate (1.30-1.41 ns/printing, 2.7-2.9x the range rate) from the same three `otag:` queries. Together the two commits flip real routing: all three now correctly pick `StreamedSelect` over `PrintingCompose`, both plans predicted within 0.90-1.08x of measured.

## Verification (re-run standalone against current `main`)
- `cargo test --release`: 157 passed, 0 failed
- `cargo clippy --release --all-targets`: clean (one pre-existing unrelated dead-code warning)

Bumps `ARCHIVE_FORMAT_VERSION` (new `CardIndexes` field `edhrec_walk_checkpoints`). If another open split also bumps it and merges first, this one needs a trivial rebase to the next number.

### DIFF
--- a/card_engine/src/cost.rs
+++ b/card_engine/src/cost.rs
@@ -207,6 +207,11 @@ pub(crate) struct PlanFeatures {
     /// uniformly across the whole permutation. `None` everywhere except a bare card-space collection
     /// leaf ordered by EDHREC, which is the only dimension `WalkCheckpoints` covers today.
     pub walked_hint: Option<f64>,
+    /// Printings a CARD-SPACE collection leaf's build broadcasts (`ids_of` +
+    /// `broadcast_card_ids_to_printings`). See `ComposeEstimate::collection_broadcast`'s doc for why
+    /// this is not just folded into `scatter_printings`. `0` for everything except `PrintingCompose`
+    /// on a card-space `subtypes`/`keywords`/`oracle_tags` leaf.
+    pub collection_broadcast_printings: u32,
 }
 
 // ─── P1: PrintingRangeScan ──────────────────────────────────────────────────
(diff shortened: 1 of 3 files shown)


jbylund:
Split into two independently-reviewable/mergeable PRs: #1023 (walk-length checkpoints) and #1024 (collection-leaf compose-build rate) — turns out they're genuinely independent code, not just independently-motivated findings. Nothing unique left here.

PR #1023, opened 2026-08-25
Engine: Exact Walk-Length Checkpoints for otag:/keyword:/t: Ordered by EDHREC

jbylund:
Split from #1017 into two independently-reviewable fixes — see that PR's description for full context. This is the first of the two; the second (a separate, unrelated pricing bug found while validating this one) is #1017's other commit, split out separately.

`docs/issues/local-engine-compose-paging-cost-based.md` leaves the Perm/OrderbyWalk walk-length estimate ("how many printings does the walk examine to fill a page") as a single global rate, and says plainly that "how matches clump along a sort order is not something a density ratio can see, and no constant will fix that." `otag:triggered-ability` measured exactly that: the formula predicted 602 printings walked against a real 1,942 — a 3.2x miss a ~10% printings-per-matching-card skew cannot explain, so the error is EDHREC-order clumping specific to this value, not a rate problem.

`WalkCheckpoints` replaces the single rate with five exact anchors per (value, direction): the permutation position at which cumulative PRINTINGS from matching cards reaches 1/100/200/400/800 (not card count — a card-level match can contribute many printings at once, so the threshold has to be printing-weighted). Interpolated at query time between whichever two anchors bracket `offset+limit`. 20 bytes per (value, direction), built once at store time from data already computed for other purposes.

Scoped to `unique=printing` ordered by EDHREC, `subtypes`/`keywords`/`oracle_tags` only: the dominant orderby (40% of `REALISTIC_ORDERBY_WEIGHTS`) and exactly the fields producing every `otag:`-shaped mis-route measured this session. `art_tags`/`is_tags` are printing-space (no card permutation to build against); card/artwork mode and the other 7 orderby columns are unattempted.

Verified against the real corpus: `printings_walked` now predicts 1,998 against a realized 1,942 (2.9% off, down from the old formula's 602 — a 3.2x miss). The mechanism is real: only 50 cards need visiting to accumulate 1,942 printings, because EDHREC's most popular cards are disproportionately old, heavily-reprinted staples (~39 printings/card among the first 50, against a 3.1 corpus average) — genuine clumping, not noise.

**Note on scope**: this fix alone barely moves `predicted_ns` for the `otag:` misroutes it was found from — the walk-length term is a small share of `PrintingCompose`'s total cost, and a second, larger pricing error (fixed in the other split) was dominating. Fixing routing for those specific queries needs both; this PR is a real, independently-verified feature-accuracy improvement on its own regardless.

## Test plan

- [x] `cargo test --release` (card_engine): 158/158 passing
- [x] `cargo clippy --release --all-targets`: clean (one pre-existing unrelated dead-code warning)

Bumps `ARCHIVE_FORMAT_VERSION` (new `CardIndexes` field `edhrec_walk_checkpoints`) — rebased to 2026082501 since `main`'s `layout:` field already claimed 2026082302.

### DIFF
--- a/card_engine/src/cost.rs
+++ b/card_engine/src/cost.rs
@@ -210,6 +210,11 @@ pub(crate) struct PlanFeatures {
     /// in ~`page_span/selectivity` steps), while the permutation-free gather visits every match — so
     /// the formula branches on this rather than assuming one. Ignored by every other plan.
     pub compose_paging: super::ComposePaging,
+    /// Interpolated walk length from `WalkCheckpoints`, when the acquire branch found one for this
+    /// value/sort-column -- `printings_walked` uses this directly instead of assuming matches spread
+    /// uniformly across the whole permutation. `None` everywhere except a bare card-space collection
+    /// leaf ordered by EDHREC, which is the only dimension `WalkCheckpoints` covers today.
+    pub walked_hint: Option<f64>,
 }
 
 // ─── P1: PrintingRangeScan ──────────────────────────────────────────────────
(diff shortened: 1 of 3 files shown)


jbylund:
Holding this rather than merging — it's overly specialized (hardcoded to unique=printing + EDHREC + subtypes/keywords/oracle_tags; covering the other 7 orderby columns and card/artwork mode would mean a separate precomputed checkpoint table per mode × orderby × leaf-type combination, each with its own ARCHIVE_FORMAT_VERSION-gated build step) and likely to be obsoleted by a more general approach before that combinatorial buildout would even be worth doing.

The sigma decision rule work (#1009) already has a candidate replacement: uniform_mean/nhg_variance, a closed-form no-clumping estimate of the same quantity (how far into the permutation the walk goes), needing zero precomputed per-value data — just matches/n_cards/offset, which the router already has for any query. Porting that to Rust (step 3 of the sigma doc's 7-step landing plan) and using it for routing, paired with the three-phase walk as an execution-time safety net for when the no-clumping assumption is wrong, would plausibly fix the same otag:/keyword:/t: misroute this PR targets — and every other clumped leaf/order combination — without adding new on-disk structure.

Not resurrecting this unless that more general path turns out not to pan out.

PR #1056, opened 2026-08-27
Rust Perf Audit: Five Measured Fixes In Card Engine And Shared Cache

jbylund:
## Summary

Five independently-measured performance fixes from a Rust perf audit of `card_engine` and `shared_cache`, each its own commit with a kernel micro-benchmark run before and after.

- Shared cache: hash the key and sample the body once per write instead of twice (`fast_check` + `set` recomputed the same hashes) — measured 51-53% reduction on the isolated hashing cost, stable across 3 runs.
- Legalities field: cache the sorted format registry instead of cloning and sorting it on every output row — measured 98-99% reduction (~600-700ns/row down to ~10.5ns), the largest win in this set.
- `PreparedCandidates::card_ids`: replace `Box<dyn ExactSizeIterator>` with a concrete 2-variant enum so P3/P4's per-card match loops call a monomorphized `next()` instead of going through a vtable — measured ~1.1-1.2ns/card after fixing a flawed first benchmark attempt that let LLVM devirtualize the box away entirely.
- Shared cache: defer `try_lock`'s clock read to the contended path only, and cache the pid via a process-lifetime `OnceLock` instead of a syscall per call — measured 74-75% reduction on the uncontended lock/unlock cycle.
- `Match`: drop a third sort key (`prefer_score`) that every real comparator already ignored, shrinking the type from 32 to 16 bytes. Verified the key was dead two ways — a static trace of every consumer, then an empirical check (zero it and run the full 160-test suite) — before removing it. Measured 1.3-1.4x faster `select_nth_unstable_by` at the buffer sizes `select_page`/`prune_to_smallest` actually run at.

One additional candidate (`plan_cost`'s unconditional `printings_walked` call) was measured and declined: isolated cost ~0.4ns/call, far below anything a routing-decision benchmark could resolve.

Four more candidates from the same audit are documented but not implemented — see the untracked `docs/issues/local-rust-perf-audit-remaining-findings.md` on the author's machine (a body-copy reduction in `gen_cache` that touches the cross-process wire format, plus three lower-priority/reload-time-only items).

## Test plan

- [x] `cargo test --release` passes (160 passed, 0 failed) in both `card_engine` and `shared_cache` after every commit
- [x] `cargo clippy --release` clean after every commit
- [x] Each fix has a dedicated `#[ignore]`d kernel micro-benchmark (`cargo test --release bench_<name> -- --ignored --nocapture`) run 3x before and after, included in the commit message
- [x] No behavior change intended in any commit; the `Match` layout change was additionally verified by zeroing the removed key and re-running the full suite before touching the type

### DIFF
--- a/card_engine/src/bench_card_ids_dispatch.rs
+++ b/card_engine/src/bench_card_ids_dispatch.rs
@@ -1,0 +1,86 @@
+//! Micro-benchmark for perf-audit finding #1: `PreparedCandidates::card_ids` returned
+//! `Box<dyn ExactSizeIterator<Item = u32>>`, so every `.next()` in P3's and P4's per-card match
+//! loops (`run_query_streamed`, `exec_gathered_scan` — up to `n_cards` iterations each) went
+//! through a vtable instead of a monomorphized call the compiler can inline.
+//!
+//! `bench_iter_dispatch.rs` measured the same class of overhead for `run_query`'s card_ids
+//! before the `PreparedCandidates` split; that call site has since moved (this module targets
+//! the current one) and its benchmark needs a corpus store this repo doesn't ship, so this one
+//! is self-contained — synthetic ids, no archive.
+//!
+//! **First attempt at this benchmark was wrong.** Constructing the `Box<dyn ...>` and consuming
+//! it in the same closure let LLVM see straight through the allocation and devirtualize both
+//! arms identically -- the "boxed" arm came back tied with, and sometimes faster than, the
+//! concrete one, which is not physically possible (the boxed arm does strictly more work: a heap
+//! allocation plus indirect calls). In production the box crosses a real function boundary
+//! (`prep.card_ids(ctx)` is built in one function, `run_query_streamed`/`exec_gathered_scan`
+//! consume it in another, neither trivially inlined into the other), so this version forces the
+//! same shape with `#[inline(never)]` consumer functions that take the iterator as an opaque
+//! parameter, matching bench_iter_dispatch's cross-function pattern.
+//!
+//! Two shapes, matching the two arms of `CardIdIter`: a narrowed `List` (the common case once a
+//! query narrows) and a full-corpus `Range` (the `None` narrowing arm).
+//!
+//!     cargo test --release bench_card_ids_dispatch -- --ignored --nocapture
+
+use std::hint::black_box;
+use std::time::Instant;
+
+use super::CardIdIter;
+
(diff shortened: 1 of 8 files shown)


jbylund:
## End-to-end A/B: this PR vs main, through the real Falcon/WSGI stack

Ran a paired A/B against main using the real Falcon app (full middleware stack, real WSGI calls, real gzip/br compression) rather than calling the engine directly, driven by `QuerySampler` in realistic mode. Both builds were compiled as standalone wheels (not installed into the shared dev venv) so neither could clobber the other's `.so`, and every run order was checked in reverse to rule out thermal/sequential drift before trusting a result.

Measurement script (not included in this PR — reusable companion to `scripts/bench_query_latency_ab.py`, kept as a gist instead of adding scope to this diff): https://gist.github.com/jbylund/a055bedeee0b43e00fa5dc98fbbe173b

### Results

| Scenario | main | this PR | Δ (mean) | Verdict |
|---|---|---|---|---|
| No cache, default 9 fields | 183.8µs mean / 117.8µs median | 178.8µs / 112.0µs | -5.0 to -5.5µs (95% CI excludes 0, stable across reversed run order) | Faster, small |
| No cache, **+legalities field** | 322-325µs / 152-154µs | 279-284µs / 140-142µs | -40.8 to -43.2µs (~9% relative, stable across reversed run order) | Clearly faster |
| Caching enabled (repeat/hit path) | ~7.2-7.3µs | ~7.1µs | -0.1 to -0.3µs | No detectable difference (within same-build noise floor) |

### Notes

- `legalities` isn't in the API's 9 default output fields, so a default-fields-only benchmark almost entirely hides this PR's largest fix (the legality-registry cache, ~98% reduction per row in isolation). Requesting it explicitly surfaces a large, robust effect instead.
- The cache-enabled arm is a genuine null result, not a failure to detect one: warmups populate the cache before timed trials run, so that arm measures cache-*hit* latency, not the write path the double-hashing/`try_lock` fixes touch. Those fixes are nanosecond-scale (13-21ns, 5-9ns) against a ~7,000ns HTTP round trip — five orders of magnitude below anything an HTTP-level harness could resolve. Expected, given those fixes were already validated at the kernel-benchmark level in the commits themselves.
- No live Postgres involved: `AppContext.setup_complete` was stubbed (same pattern `api/tests/conftest.py`'s `stub_api_resource_fixture` uses) since `/search` never touches the DB past that one check.

